Issue #364, opened 2014-07-09
Crash on gist files > 1 MB


jwass:
The Gist API now just truncates file contents to 1 MB. https://developer.github.com/changes/2014-05-06-gist-api-now-truncating-large-files/

When truncated, the contents will not be valid JSON or GeoJSON. It will throw a SyntaxError on JSON.parse(). See http://geojson.io/#id=gist:/a0d1577b24482c3bdf95 for an example of this in action.

The gist response's file entry has a `truncated` flag to indicate if the data is chopped. So I think the fix is to check if the data is truncated and then make a second request to the `raw_url` field to grab the full contents.

This also breaks most uses of anonymous gist upload for [geojsonio-cli](https://github.com/mapbox/geojsonio-cli) and [geojsonio.py](https://github.com/jwass/geojsonio.py) for now until fixed.

geocentrum1:
Hi guys, any successful example which can load GEOJSON > 1MB from Gist? Unfortunatelly for me it doesnt work :-( GEOJSON layer: https://gist.github.com/geocentrum1/6162cda1f4102b232511

PR #367, opened 2014-07-24
Request raw data for gists that are truncated

jwass:
Gist files greater than 1 MB are truncated in the gist API.
https://developer.github.com/changes/2014-05-06-gist-api-now-truncating-large-files/.
Look for the truncated flag and request the full contents from the
raw_url if it's set. 

Addresses #364.

A few days ago this fix didn't work - gist raw files are served from gist.githubusercontent.com, which was not CORS enabled. But after an email, @github support just set `Access-Control-Allow-Origin: *`. So this works now.

### DIFF
--- a/dist/site.js
+++ b/dist/site.js
@@ -1017,13 +1017,8 @@ function Buffer (subject, encoding, noZero) {
 
   var type = typeof subject
 
-  // Workaround: node's base64 implementation allows for non-padded strings
-  // while base64-js does not.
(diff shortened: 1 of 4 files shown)


geocentrum1:
Hi guys, any successful example which can load GEOJSON > 1MB from Gist? Unfortunatelly for me it doesnt work :-( GEOJSON layer: https://gist.github.com/geocentrum1/6162cda1f4102b232511

PR #398, opened 2014-12-30
Fixes gist URL API

jwass:
Adds a missing '/' in the URL used to fetch a gist.

For example, here's the link in API.md which is currently broken:
http://geojson.io/#id=gist:tmcw/e9a29ad54dbaa83dee08&map=8/39.198/-76.981

Previously was trying to fetch URLs like: https://api.github.com/gistsabcde - which would 404.
Proper URL is like: https://api.github.com/gists/abcde

### DIFF
--- a/dist/site.js
+++ b/dist/site.js
@@ -24202,7 +24202,7 @@ function save(context, callback) {
 }
 
 function load(id, context, callback) {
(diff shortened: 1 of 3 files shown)


tmcw:
:+1: thanks @jwass !

Issue #232, opened 2013-09-08
Add wkt/wkb support


sabas:
Could load directly a result of a postgis query via exporting in csv, example

```
SELECT osm_id,geometry FROM planet_osm_point;
```

Bluesky again :-)

shane-axiom:
Are the devs still opposed to WKT support? I see that there [used to be opposition to supporting shapefiles](https://github.com/mapbox/geojson.io/pull/242), but I see that shapefile is now an export option in the save menu. WKT export seems like another natural export option, even if import isn't supported. And yes, Wicket is awesome, but it would be nice to have all this functionality in one place.

tmcw:
> I see that there used to be opposition to supporting shapefiles, but shapefile is now an export option in the save menu.

Exporting Shapefiles is reasonable, importing the is not: it's at least 10x more complex to support.

Problem is as previously stated: this would require a new UI. If anyone wants to contribute that UI, go for it.

shane-axiom:
@tmcw Just to clarify, I'm not asking for shapefile import. I'm asking for WKT export.

tmcw:
Yes, just to be clear about my reasoning on the other issue, which was a rejected PR for shapefile importing, and I eventually implemented exporting. Which is to say: i'm more liberal with adding export formats than import formats, because you can safely export one kind of thing, but in importing you need to handle all variants.

PR #399, opened 2015-01-03
Add WKT download option

shane-axiom:
Adds WKT download/export option using https://github.com/mapbox/wellknown. Since wellknown.stringify can't accept FeatureCollections, each feature is stringified individually and output on a separate line.

See https://github.com/mapbox/geojson.io/issues/232

Note: Based on other pull requests I'm guessing the dist files should be included. If not let me know and I can remove them.


src/ui/file_bar.js
```diff
@@ -448,6 +452,21 @@ module.exports = function fileBar(context) {
         }
     }
 
+    function downloadWKT() {
+        if (d3.event) d3.event.preventDefault();
+        var contentArray = [];
+        var features = context.data.get('map').features; 
+        if (features.length == 0) return;
+        features.forEach(function(feature) {
```
tmcw: This can be more succinctly expressed as

```js
var content = features.map(wellknown.stringify).join('\n');
```

shane-axiom:
Applied @tmcw's suggestion to use [Array.map](https://github.com/mapbox/geojson.io/pull/399/files#diff-4229365b4aeb78a4b3bd9fc1a94ef9e6R460).

Issue #385, opened 2014-09-23
GitHub and Gist files do not load from file browser (must reload page)


landonreed:
When selecting a gist or GitHub file from the geojson.io file browser, you must reload the page in order for the data to load.  I'm not entirely sure what the source of this error is, but does it have something to do with [`d.content` in `data.parse()`](https://github.com/mapbox/geojson.io/blob/gh-pages/src/core/data.js#L213) being undefined?

I'm getting this behavior in multiple browsers on OS X.

vappiah:
Just had the same issue

PR #400, opened 2015-01-07
Fixes #394 and #385

### LINKED ISSUE #394: Nothing happens on open from github
Maybe I am confused on how this is supposed to work, but when I click open, github, navigate to the file, it shows up to the right of the menu, but is not loaded in the json editor window and does not show up on the map. I can copy/paste the same file into the editor window, edit it and save it back to github without issues. I am looking for a way to host and edit simple map files, and have multiple contributors, and this seems like a cool way to do that, too bad it does not work.

landonreed:
Fixed fileUrl function to point to `/repos/` instead of `/gist` for GitHub repo calls

### DIFF
--- a/src/source/github.js
+++ b/src/source/github.js
@@ -115,7 +115,7 @@ function loadRaw(parts, sha, context, callback) {
 }
 
 function fileUrl(parts) {
-    return githubBase + '/gists' +
+    return githubBase + '/repos/' +
         parts.user +
         '/' + parts.repo +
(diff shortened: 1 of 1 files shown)


landonreed:
I think this may fix some of the refactoring repo issues @tmcw and @camilleanne?

camilleanne:
Seems to be part of it. I'll roll this into some of the other bug fixes I'm working on.

camilleanne:
pulled into #401

Issue #415, opened 2015-02-17
popup too big


samanpwbb:
with https://github.com/mapbox/geojson.io/pull/411 popups got pretty unwieldy:

![screen shot 2015-02-17 at 11 16 21 am](https://cloud.githubusercontent.com/assets/108094/6233042/35ad9574-b69e-11e4-9f3d-f4439c95a75e.png)

When trying to draw data, each click forces map to pan to fit the new popup and it's pretty disorienting and slows down the workflow. 

- Popups should have flexible height.
- Popups should use space more efficiently and hide UI components that aren't relevant. 
- Condense all the margins/paddings.

@bmcbride do you have time to have a go at this?

bmcbride:
@samanpwbb What size screen are you on? You can mess around with resizing the popup in the CSS here: https://github.com/mapbox/geojson.io/blob/gh-pages/css/marker.css#L47-L58. What size would you prefer?

samanpwbb:
I'll have a go at this :stuck_out_tongue:

PR #416, opened 2015-02-19
fix popup styling

samanpwbb:
Looks like this:
3a37aa14-b7aa-11e4-991e-0f8fbb9a434f copy

Fixes https://github.com/mapbox/geojson.io/issues/415

cc: @tmcw


tmcw:
:+1: LGTM!

PR #422, opened 2015-02-20
Allow users to easily add custom tile layers

bmcbride:


tmcw:
:+1: awesome, thanks!

PR #427, opened 2015-02-27
Update share.html to Mapbox.js 2.1.4

### LINKED ISSUE #403: Update mapbox.js to 2.1.4 on shared maps
Geojson.io is mostly updated to Mapbox.js 2.1.4 with the exception of [share.html](https://github.com/mapbox/geojson.io/blob/gh-pages/data/share.html#L32-L34). This should be updated at some point from 1.6.2, `detectRetina: true` removed, and `access_token` from [config.js](https://github.com/mapbox/geojson.io/blob/gh-pages/src/config.js#L3) added.

snkashis:
Fixes #403

### DIFF
--- a/data/share.html
+++ b/data/share.html
@@ -29,9 +29,9 @@
     background-color:#f7f7f7;
 }
   </style>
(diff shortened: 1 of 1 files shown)


tmcw:
Superseded by #477

PR #435, opened 2015-04-01
Maintain editor scroll position after map change

davebrent:
This fixes a bug where, if you have scrolled down the JSON panel, made an edit, then added added a new feature on the map, the JSON panel would scroll back to the top instead of maintaining its position


tmcw:
:+1: awesome, thanks!

PR #436, opened 2015-04-10
use geojson-normalize to normalize inputs

### LINKED ISSUE #352: Open->File does not handle non-featurecollections
When opening GeoJSON file from local source through the menus, file does not load. Dragging and dropping does work.

jseppi:
this allows for import of GeoJSON features and geometries
fixes #352


tmcw:
:+1: merged & merged in #451

PR #444, opened 2015-05-18
Add OpenCycleMap option to layer switcher

mvl22:
OpenCycleMap is a very popular OSM tileset. We would find this a very useful background to have available.

### DIFF
--- a/src/ui/layer_switch.js
+++ b/src/ui/layer_switch.js
@@ -23,6 +23,11 @@ module.exports = function(context) {
                 title: 'Satellite',
                 layer: L.mapbox.tileLayer('tmcw.map-j5fsp01s')
             }, {
+                title: 'OCM',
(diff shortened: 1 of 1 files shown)


mvl22:
NB I have made contact with the maintainer of the OpenCycleMap server whose bandwidth would be used, @gravitystorm, and he has no objections.

dmofot:
Just an FYI, but you can add any tileset you want through Meta -> Add map layer.  I just added [OCM](http://{s}.tile.thunderforest.com/cycle/{z}/{x}/{y}.png) and it worked flawlessly and adds a little checkbox to turn it on/off down in the lower right corner.  Probably a better option that trying to accommodate every tileset in the switcher.

image

tmcw:
:+1:

mvl22:
Cheers :)

PR #454, opened 2015-06-15
Swap attributionControl for infoControl

bmcbride:
The attribution is currently covering the basemap buttons on smaller screens. This frees up some room, allowing for more basemaps and attribution/info.


tmcw:
:+1:

PR #477, opened 2015-09-17
pull in mapbox.js 2.2.2

tcql:
updates the local copy of mapbox.js & resources to 2.2.2

/cc @tmcw


tcql:
updated to include #403 / #427

Issue #489, opened 2015-12-23
Improve the map link goes to null island


planemad:
Go here http://geojson.io/#map=6/21.923/-83.650 don't interact with the map. Click **improve the map**, the coordinates of the current location is not passed.

tmcw:
I think this is upstream in Mapbox.js

tmcw:
This was fixed in #494 (thanks @jwass !)

jwass:
@tmcw and thanks to you for taking a look and pushing live quickly.

PR #494, opened 2016-02-06
Use attribution control

### LINKED ISSUE #482: replace InfoControl with default attribution control
InfoControl is [deprecated](https://github.com/mapbox/mapbox.js/issues/941).

jwass:
Closes #482 and #489 

Disables the infoControl. The previous pieces of info (Feedback
and About) are now set as the prefix in the attribution control.

P.S. I'm never quite sure if I'm handling site.js and site.mobile.js properly. After `npm install` and `make` it seems to write a completely different file despite only changing a few lines.

### DIFF
--- a/dist/site.js
+++ b/dist/site.js
@@ -1234,6 +1234,8 @@ module.exports=[
 ]
 
 },{}],2:[function(require,module,exports){
+
+},{}],3:[function(require,module,exports){
 // http://wiki.commonjs.org/wiki/Unit_Testing/1.0
 //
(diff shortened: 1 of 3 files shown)


tmcw:
:+1: awesome. The random re-sorting of items is because of two things - without an `npm-shrinkwrap`, everyone installs a slightly different version of dependencies, and since browserify isn't deterministic, it doesn't always produce the same bundle. Both of these are kind of deep-seated isses in JavaScript-land.

Issue #499, opened 2016-02-23
allow more external sources through Content Security Policy?


drewda:
@tmcw are you open to loosening CSP somewhat? 

After #496, queries like http://geojson.io/#data=data:text/x-url,https%3A%2F%2Ftransit.land%2Fapi%2Fv1%2Foperators.geojson are no longer possible--even though the target URL is serving CORS headers.

![geojson_io](https://cloud.githubusercontent.com/assets/212369/13265729/63934d1c-da29-11e5-9f8b-de729b54934c.png)

tmcw:
Yep, I've been ratcheting things toward the 'much more secure side' and they could maybe move a step or two backward.

drewda:
:+1: thanks, @tmcw.

tmcw:
Loosened connect-src restrictions in 059c105ed960c67d056e7a1c7a4f9d51b818f34f

bmcbride:
@tmcw Any chance of loosening the restrictions for img-src as well? You can no longer add custom basemaps via Meta > Add map layer. Thanks!

bmcbride:
Thanks!

PR #513, opened 2016-05-09
Eliminate CSP warning for data: scheme

bhousel:
Need to explicitly add `data:` to the content security policy to avoid warnings on `data:` urls
`*` does not catch these, because it's treated as a source, not a protocol
see also http://stackoverflow.com/a/18449556/7620

relates to #499

### DIFF
--- a/index.html
+++ b/index.html
@@ -44,6 +44,7 @@
       ;
     img-src
       'self'
+      data:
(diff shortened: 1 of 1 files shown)


tmcw:
👍 thx

Issue #485, opened 2015-11-12
Public link does not display colors I set using the site


owen-dall:
http://bl.ocks.org/d/52b6cbc7e5c17337ab7a

See screen shot of how it looks on geojson.io.
![buildings with color](https://cloud.githubusercontent.com/assets/10927/11117274/6f2275d2-8906-11e5-8cf9-2a51d61f063c.png)

tmcw:
Indeed, for some reason we aren't using L.mapbox.featureLayer in the embed.

PR #519, opened 2016-06-20
Fix for #485 - Updating embed HTML to use L.mapbox.featureLayer

davidchouse:
Updating embed HTML to use Mapbox FeatureLayer to support icon styling.


tmcw:
👍 thanks david!

PR #525, opened 2016-08-03
Set maxZoom on custom map layer to current map maxZoom. Fixes #521

### LINKED ISSUE #521: No tiles loaded from custom map layer when zoom >18
I have added a custom map layer to the map using the menu. The map layer supports up to zoom level 20. When I zoom past zoom level 18 the layer disappears, as if geojson.io isn't requesting the tiles past this zoom level. How can I request the tiles for my custom layer past zoom level 18?

andrewharvey:


tmcw:
👍 Thanks!

Issue #526, opened 2016-08-12
Integration with geofenceapi.org


blockjon:
GeofenceAPI.org is a hosted geofence engine based on GeoJSON.

I was interested in building a new "Save" option in order to have an ability to save my GeoJSON file into https://api.geofenceapi.org/ ... If I put an upstream PR together with this new "Save" option, will it get accepted?

Thank you,
Jon

tmcw:
Hey Jon,

Sure, if it's of the same style as the save-to-Gist option, I'd accept that PR.

- Tom

blockjon:
https://github.com/mapbox/geojson.io/pull/527

tmcw:
Fixed in #527

PR #527, opened 2016-08-16
Geofenceapi.org integration

blockjon:
Here's my first crack at integrating with GeofenceAPI.org.

The idea here is that someone who's creating a GeoJSON document can import their work directly into GeofenceAPI.org in just one step.

geojsonio

I have intentionally only done a small amount of work here to demonstrate the integration and solicit feedback.

@tmcw in reference to https://github.com/mapbox/geojson.io/issues/526


src/source/geofenceapi.js
```diff
@@ -0,0 +1,38 @@
+var fs = require('fs');
```
tmcw: looks like this and the `tmpl` variable aren't used

tmcw:
besides the inline comment about unused vars, this looks pretty good!

blockjon:
@tmcw Per your comment, I just removed what i think are more unused lines via https://github.com/mapbox/geojson.io/pull/527/commits/3be77a9e08587cd7c7b0ce4d06856d116caf7a66

On my end, as I move forward, I will "harden" the landing page I created today unit & integration tests and also streamline the user's experience one they land.

At this stage, I will commit to being extremely careful to keep it working and available via my chatroom tool and/or github if someone reports a glitch.

What else can I provide to you in advance of us merging & deploying?

blockjon:
It's ready @tmcw

tmcw:
👍 merging, thanks!

blockjon:
🎉

blockjon:
@tmcw I'm shutting the service down since it hasn't proven viable to keep running. I tried to create a revert but GH prevented it. Let's knock the link out for Geofenceapi.

Issue #459, opened 2015-06-18
Unicode decoding issue when loading file from GitHub


ronnix:
When uploading geoJSON file from my computer, properties that include non-ASCII Unicode characters display correctly (e.g. "Café").

When loading the exact same file from GitHub, properties are not displayed correctly (e.g. "CafÃ©").

Note that this happens whether the file is UTF-8 encoded, or if I use `\u00e9` notation for non-ASCII characters.

jeekajoo:
Same here.
https://github.com/ParisEnSelle/planvelo-carte/blob/master/planvelo.geojson
https://github.com/ParisEnSelle/planvelo-carte/blob/master/planvelo_modified.geojson

Diff extract
```
311c311
<         "name": "Place des Fêtes"
---
>         "name": "Place des FÃªtes"
```

jeekajoo:
```bash
$ file planvelo*
planvelo.geojson:          UTF-8 Unicode text
planvelo_modified.geojson: UTF-8 Unicode text
```
@tmcw

Loufylouf:
This is apparently a well known problem that is documented here : https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding#The_Unicode_Problem . I tried the first solution which works perfectly, though I'm not sure that it is the most efficient. 

If you wish to integrate it, I suggest modifying src/ui/file_bar.js (lines 290-298) from 
```
                    if (last.type === 'blob') {
                        githubBrowser.request('/repos/' + d[1].full_name +
                            '/git/blobs/' + last.sha, function(err, blob) {
                                d.content = JSON.parse(atob(blob[0].content));
                                context.data.parse(d);
                                zoomextent(context);
                                m.close();
                            });
                    }
```
to : 
```
                    if (last.type === 'blob') {
                        githubBrowser.request('/repos/' + d[1].full_name +
                            '/git/blobs/' + last.sha, function(err, blob) {
                                d.content = JSON.parse(decodeURIComponent(Array.prototype.map.call(atob(blob[0].content), function(c) { return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2); }).join('')));
                                context.data.parse(d);
                                zoomextent(context);
                                m.close();
                            });
                    }
```
It's a quick & dirty fix but I'm not familiar enough with geojson.io to know where to put the b64DecodeUnicode function definition.

mykolakozyr:
[...]

PR #543, opened 2017-01-17
Quick fix to solve issue #459 (unicode decoding problem when loading …

Loufylouf:
…geojson from Github).


jeekajoo:
fixes #459 (commenting to add the reference)

Loufylouf:
I also added the possibility to update files on Github instead of having to create a new one. Should be in a separate pull request, but I can't find how to do that.

tmcw:
Thanks, merged!

Loufylouf:
You're welcome. I haven't updated the generated files site.js and site_mobile.js, because, well they should be generated instead of taken straight from Github, but I can understand that for an easy distribution it might be better, so you might have to add a commit with the content of the files updated.

jeekajoo:
Thanks much to both of you!

PR #548, opened 2017-02-13
make sure generated polygons follow right-hand-rule

tyrasd:
to comply with https://tools.ietf.org/html/rfc7946#section-3.1.6


tmcw:
Thanks Martin!

PR #550, opened 2017-02-16
Revert "rotate access token"

tmcw:
Reverts mapbox/geojson.io#549

### DIFF
--- a/data/share.html
+++ b/data/share.html
@@ -36,7 +36,7 @@
 <body>
 <div id='map'></div>
 <script type='text/javascript'>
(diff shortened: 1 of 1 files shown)


tomscholz:
why?

tmcw:
I reverted this revert. Basically, the token got cycled, the site broke, and I assumed it was the wrong token it got cycled to. That assumption was wrong - the new token is correct, but it was only changed for the share page, not geojson.io's main site itself.

tomscholz:
ok ok :D

PR #557, opened 2017-03-14
WKX (WKT/WKB) support

lokkju:
This small change adds three menu items to the Meta menu, allowing the import and conversion of WKT strings, hex encoded WKB strings, and base64 encoded WKB strings.  It cleanly adds the basic features requested by issue #232, without dealing with file format issues.

The added libraries, WKX and Buffer, are both MIT licensed

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,6 @@ This is made up of small reusable chunks:
 * <a href='https://github.com/tyrasd/osmtogeojson'>tyrasd/osmtogeojson</a>
 * <a href='https://github.com/w8r/Leaflet.draw.drag'>w8r/Leaflet.draw.drag</a>
 * <a href='https://github.com/cschwarz/wkx'>cschwarz/wkx</a>
-* <a href='https://github.com/feross/buffer'>feross/buffer</a>
 
 
 ## See Also
--- a/README.md
+++ b/README.md
@@ -60,6 +60,9 @@ This is made up of small reusable chunks:
 * <a href='http://github.com/mapbox/togeojson'>mapbox/togeojson</a>
 * <a href='https://github.com/tyrasd/osmtogeojson'>tyrasd/osmtogeojson</a>
 * <a href='https://github.com/w8r/Leaflet.draw.drag'>w8r/Leaflet.draw.drag</a>
+* <a href='https://github.com/cschwarz/wkx'>cschwarz/wkx</a>
+* <a href='https://github.com/feross/buffer'>feross/buffer</a>
+
 
 ## See Also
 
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
     "@mapbox/github-file-browser": "0.6.1",
     "bowser": "0.3.2",
     "brfs": "^1.4.3",
+    "buffer": "^5.0.5",
```
tmcw: you don't need to declare buffer as a dependency. as [its readme makes clear](https://github.com/feross/buffer#buffer---), it is automatically included with browserify.

src/lib/meta.js
```diff
@@ -3,6 +3,8 @@ var escape = require('escape-html'),
     geojsonExtent = require('geojson-extent'),
     geojsonFlatten = require('geojson-flatten'),
     polyline = require('polyline'),
+    wkx = require('wkx'),
+    Buffer = require('buffer/').Buffer,
```
tmcw: You don't need to require Buffer - browserify will add the global when you refer to it.

src/lib/meta.js
```diff
@@ -52,3 +54,39 @@ module.exports.polyline = function(context) {
...
+
+module.exports.wkxHex = function(context) {
+    var input = prompt('Enter your Hex encoded WKB/EWKB');
+    try {
+        var decoded = wkx.Geometry.parse(Buffer.from(input,'hex'));
+        context.data.set({ map: decoded.toGeoJSON() });
+        zoomextent(context);
+    } catch(e) {
+        console.error(e)
+        alert('Sorry, we were unable to decode that Base64 encoded WKX data');
```
tmcw: Says Base64 encoded here, should say Hex encoded.

lokkju: Think you put your comment in the wrong place.  This block is base64; hex is a different block, and did need fixed.  it now is.

README.md
```diff
@@ -60,6 +60,9 @@ This is made up of small reusable chunks:
 * <a href='http://github.com/mapbox/togeojson'>mapbox/togeojson</a>
 * <a href='https://github.com/tyrasd/osmtogeojson'>tyrasd/osmtogeojson</a>
 * <a href='https://github.com/w8r/Leaflet.draw.drag'>w8r/Leaflet.draw.drag</a>
+* <a href='https://github.com/cschwarz/wkx'>cschwarz/wkx</a>
+* <a href='https://github.com/feross/buffer'>feross/buffer</a>
```
tmcw: Remove Buffer from here too

tmcw:
Thanks!

Issue #565, opened 2017-04-28
Disable drag to move polygon while editing.


jolyonralph:
It is HUGELY annoying when editing points on a complex polygon when I accidentally miss one point and instead start to move the entire polygon around.   I can't think of a single time, ever, that I'd want to move a polygon around the map compared to the times I'd want it to stay exactly where it is. 

Can we please, please, please have an option to disable the ability to drag polygons around while editing. I want to manipulate the points, nothing else.

Thank you :)

jolyonralph:
Please? Anyone? This is driving me crazy.

jolyonralph:
Keeps happening. Still infuriating.

Is this project still supported? Just curious.  Are there any alternatives?

tomscholz:
I implemented this feature, but I don't see any way to do, what you are looking for. You could clone the repo and remove the feature, If you don't need the moving feature.

jolyonralph:
Can we PLEASE at least have an option to UNDO a move?  

Honestly, adding this feature without having a way to disable or undo it was pretty reckless.

thisislawatts:
Hey @jolyonralph, you could try out a site like scribblemaps.com they have a pretty sophisticated editing tool and the free version should do most things you need. Although their paid plan is pretty reasonable if you're not looking to make a PR on this repo to implement the features you're after.

jolyonralph:
Thanks, I will check this out. 

Although I am frustrated because geojson.io does everything that I need perfectly *except* this awful move issue which has cost me countless time.

I would not be opposed to paying someone here to fix this bug/"feature" if the price was fair.

jolyonralph:
Unfortunately scribblemaps.com isn't appropriate. I'm trying to encourage people within our community (http://www.mindat.org) to generate geojson maps of areas of geological/mineralogical interest (eg mining districts.)   I can't ask everyone to pay $99 a year for that.  But I don't mind paying a little for someone to add the required option to disable drag-to-move here.

jolyonralph:
ps. my preferred solution to this issue looking at how the UI is currently built is to change the system so that to drag an existing polygon around a map you need to hold shift while dragging with the mouse.

tomscholz:
Shift is already in use to zoom

thisislawatts:
I believe that scribblemaps might have a package which allows you to use it as part of a service, but understand your point. I ultimately switched to using geojson.io because of it's focused feature set.

[...]

PR #568, opened 2017-05-13
Use :before pseudo element to increase hit size of points

thisislawatts:
This makes it easier to drag and drop points, rather than entire shapes.

__Gif demonstrating markers with larger hit targets__ 

geojsondraggablemarkers

__Gif demonstrating larger hit targets and :hover styling to emphasis active marker__  

geojsondraggablemarkerswithoutlines

Ref: #565

### DIFF
--- a/css/marker.css
+++ b/css/marker.css
@@ -111,3 +111,28 @@ datalist {
     cursor: pointer;
     color: #199CF4;
 }
+
+
+/* Editable Points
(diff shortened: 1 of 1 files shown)


thisislawatts:
Oops, left some debug outlines in there. Will tidy this up and resubmit next week. Apologies for the noise 👁  Watchers

Issue #573, opened 2017-06-23
Opening Gists > 1 MB fails


bealeofsteel:
When a Gist is greater than 1 MB in size, Github returns a truncated version of it, causing JSON parsing to fail.  Truncated files should be fetched in full from the raw URL.  

Full error: 

`SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at Object.data.parse (VM158 site.js:28454)
    at VM158 site.js:30306
    at VM158 site.js:1297
    at Object.callback (VM158 site.js:1328)
    at on_end (VM158 site.js:2371)
    at XMLHttpRequest.on_state_change (VM158 site.js:2321)`

bhousel:
related:  https://github.com/mapbox/geojsonio-cli/issues/13

j3j5:
Just to add here, I don't know enough JS to make a full MR but I was able to track down the problem to the following.

Clicking Open -> Gist runs [this code](https://github.com/mapbox/geojson.io/blob/eced11bc78acce0803af3311a67c34891c4cce6a/src/ui/file_bar.js#L420) which sends the result of the API response to `data.parse()`, but at that point the content is truncated for files > 1MB, so my guess is that on [this catch](https://github.com/mapbox/geojson.io/blob/eced11bc78acce0803af3311a67c34891c4cce6a/src/core/data.js#L254) another call to `source.gist.loadRaw()` should be made, just like it's done on `data.fetch()` ([a few lines above](https://github.com/mapbox/geojson.io/blob/eced11bc78acce0803af3311a67c34891c4cce6a/src/core/data.js#L120)). Actually, the ideal would be to do just like there and check whether the file has the `truncated` flag set to `true`, and if so call for the raw version.

Hope this helps!

chriswhong:
Closing as no longer relevant as gist support has been removed.

PR #576, opened 2017-07-14
Fix for Gists failing to load when >1MB in size

bealeofsteel:
Currently attempting to load a Gist >1MB in size fails, as detailed in #573.  This addresses the issue by fetching the full contents of the Gist from the raw_url.

### DIFF
--- a/index.html
+++ b/index.html
@@ -24,6 +24,7 @@
   <![endif]-->
   <script src='lib/mapbox.js/latest/mapbox.js'></script>
   <script src='lib/raven.min.js'></script>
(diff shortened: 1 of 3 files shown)


ingalls:
Would love to merge this!

I would need you to make a few changes however. The biggest one is that JQuery is a massive library that is mostly convenience functions in modern web browsers.

If you were to refactor this code to use native  [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) I would be happy to merge.

PR #578, opened 2017-07-28
add HERE Maps API to list of map data provider

andirotter:
adds the HERE Maps API to the bottom left menu and retrieves the map tiles from there


src/ui/layer_switch.js
```diff
@@ -32,6 +32,11 @@ module.exports = function(context) {
                 layer: L.tileLayer('https://a.tile.openstreetmap.org/{z}/{x}/{y}.png', {
                     attribution: '&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
                 })
+            }, {
+                title: 'HERE',
+                layer: L.tileLayer('http://1.base.maps.cit.api.here.com/maptile/2.1/maptile/newest/normal.day/{z}/{x}/{y}/256/png8?app_id=xWVIueSv6JL0aJ5xqTxb&app_code=djPZyynKsbTjIUDOBcHZ2g', {
```
jieter: From which account is this `app_id`?

timaschew: Looks like from `view-source:https://wego.here.com/` (paste it in chrome)
line 128

ingalls:
I'm afraid the HERE terms of use are unnecessarily restrictive.

> GIS (geographic information systems) applications, which means an application that analyzes, processes or visualizes (or any combination of them) information in order to manage land use, natural resources, environment, urban or enterprise facilities, communications or utility networks or any other fixed assets; 

Custom Layers can be added using the 

> Meta > add map layer

Option.  Allowing user to add any number of tile providers.

PR #588, opened 2017-12-03
updated d3 to v4

simonseyock:
Hi!

I am working on a modified version of geojson.io at the moment and wanted to upgrade d3 to version 4. Maybe you are interested in this.

There are 3 points left which would need some work / discussion:
1. I was not able to test github and gist as I was not able to login from my local machine (probably missing some github/gatekeeper api stuff).
1. I switched from importing only needed methods to importing only needed modules. This decreased the size of the lib.js, though.
1. The icons from the d3-metatable buttons are missing.

If you are interested I can put some more work into it and try to solve this issues.


simonseyock:
One motivation for the upgrade is that d3 v3 dependents on a package that is not possible to install with node version > ~6.7

Tashaa78:
Hie, am a newbie, how do i build this project, i wanted the updates so i took your commit as the starting point.

sgillies:
Per the note at the top of the readme, we're only fixing critical bugs at this time. The issue with the node version is close, but will not be addressed.

PR #592, opened 2018-01-01
Export to GPX

m4ci3k2:
Adds a feature to export data to GPX format. This would be very useful as this is the only format supported for loading into OsmAnd.


sgillies:
@m4ci3k2 please see the note at the top of the readme. We're only accepting fixes for critical bugs at this point.

PR #595, opened 2018-02-14
Remove feedback link

tmcw:
The feedback link on geojson.io reports to my inbox, and I no longer maintain / have commit access / provide support for geojson.io.

Feel free to replace it with a link to a Mapbox-operated feedback mechanism, or anything else, but it's unfortunate for it to be essentially `/dev/null` right now.


bhousel:
Thanks @tmcw - I agree it would be great to replace this with something that a human will see, but for now we'll just remove it.  People can still find their way here via the About page.  We're discussing ways that we can continue to provide limited support.

PR #604, opened 2018-03-01
Disable OCM layer

### LINKED ISSUE #577: OpenCycleMap layer needs API key
The (very useful) OpenCycleMap layer is reporting that requests are not using an API key, which has been recently introduced:

screen shot 2017-07-28 at 13 53 18

sgillies:
OCM is no longer free for all. I propose to disable the layer, leaving instructions on how to enable it.

Resolves #577

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Next Release
+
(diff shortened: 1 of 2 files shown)


sgillies:
@bhousel it's still around, but requires an API key. In #577 I came to the conclusion that we'd need to pay and that we don't have a mission to provide free access to commercial tiles and no interest in charging for use of geojson.io.

PR #622, opened 2018-05-14
updated d3 to v4

tmcw:


sgillies:
@tmcw I like the looks of this! Is there anything else in #588 to be picked up? If not, I'll close that one.

tmcw:
Whoops, I meant to target the merge at tmcw/geojson.net rather than mapbox/geojson.io (GitHub targets the parent repo by default.) geojson.net is a fork that aims to fix the outstanding issues in geojson.io, address a handful of long-standing problems, and mildly modernize the codebase. basically 'keep it going' in a minimal sense.

PR #637, opened 2018-11-01
make the scroll bar wider so it's easier to see and grab

andrewharvey:


sgillies:
@andrewharvey please see the note at the top of the readme. We're only accepting fixes for critical bugs at this point.

Issue #639, opened 2018-11-16
npm install error


zyp001a:
I tried these version of node for `npm install`
* node/8.*
* node/9.11.2
* node/10.13.0
* node/11.2.0
they all failed in building process.
only node/6.14.4 works

jackyq2015:
Same issue here:
`531 verbose stack Error: contextify@0.1.15 install: `node-gyp rebuild`
531 verbose stack Exit status 1
531 verbose stack     at EventEmitter.<anonymous> (/home/cui/.nvm/versions/node/v8.15.0/lib/node_modules/npm/node_modules/npm-lifecycle/index.js:301:16)
531 verbose stack     at emitTwo (events.js:126:13)
531 verbose stack     at EventEmitter.emit (events.js:214:7)
531 verbose stack     at ChildProcess.<anonymous> (/home/cui/.nvm/versions/node/v8.15.0/lib/node_modules/npm/node_modules/npm-lifecycle/lib/spawn.js:55:14)
531 verbose stack     at emitTwo (events.js:126:13)
531 verbose stack     at ChildProcess.emit (events.js:214:7)
531 verbose stack     at maybeClose (internal/child_process.js:915:16)
531 verbose stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:209:5)
532 verbose pkgid contextify@0.1.15
533 verbose cwd /media/cui/deve/src/geojson.io
534 verbose Linux 4.15.0-43-generic
535 verbose argv "/home/cui/.nvm/versions/node/v8.15.0/bin/node" "/home/cui/.nvm/versions/node/v8.15.0/bin/npm" "install"
536 verbose node v8.15.0
537 verbose npm  v6.4.1
538 error code ELIFECYCLE
539 error errno 1
540 error contextify@0.1.15 install: `node-gyp rebuild`
540 error Exit status 1
541 error Failed at the contextify@0.1.15 install script.
541 error This is probably not a problem with npm. There is likely additional logging output above.
542 verbose exit [ 1, true ]
`
Maybe that's the reason to move to geojson.net?

sgillies:
There are no plans to fix this, but we will add an engines item to the package.json to make the situation more clear.

PR #646, opened 2019-02-25
Requires Node versions < 8

sgillies:
Node 7 was the current version at the time package.json was last modified and users report in #639 that the app doesn't install with 8+.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,5 +66,6 @@
   },
   "window": {
     "toolbar": false
-  }
+  },
+  "engines" : { "node" : "<8" }
 }


sgillies:
@bhousel yes? no?

bhousel:
From reading here, it looks like we can just drop the contextify dependency:
https://github.com/brianmcd/contextify

(I see issues like this a lot with node-pre-gyp failing on newer node builds, and in almost all cases we can find a workaround or pure javascript replacement)

sgillies:
Thanks @bhousel. Someday, perhaps. In the meanwhile, I think this will make the current state of affairs more clear, yes?

bhousel:
> Thanks @bhousel. Someday, perhaps. In the meanwhile, I think this will make the current state of affairs more clear, yes?

Looks good 👍

PR #664, opened 2020-05-12
update from Studio Classic styles ahead of deprecation

mikelmaron:
referencing new styles via `L.tilelayer` with url template because Mapbox.js 2.2.2 does not support `L.mapbox.styleLayer`, and attempt at updating led to further complications


pratikyadav:
giphy (1)

woodElec:
Hi, did those changes affect the zoom level ?

sgillies:
@woodElec yes it did, indirectly. We've fixed the regression in #667 and re-deployed.

PR #678, opened 2021-02-12
add homepage to package.json

coliff:
REF: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#homepage


sgillies:
@coliff as it says in the README we're not accepting PRs at this time.

PR #688, opened 2021-12-01
Docker container for local deployment

PeterBaker0:
Just wanted a local instance of this and ended up writing a Dockerfile - had some compatibility issues with various node versions so used node:6 for building the site and node:16 for running the http file server.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+
+node_modules
+npm-debug.log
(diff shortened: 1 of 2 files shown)


grzaks:
Very helpful. I had the same issues with node versions and this Dockerfile was the easiet way to solve them. Good job! :)

chriswhong:
Hi, thanks for working on this.  As of this morning we merged some updates that relieve the dependency on node 6.  I'm going to close this issue for now, but happy to add a Dockerfile if it's something that will help your workflows.

PR #715, opened 2022-10-20
restore add raster layer functionality

chriswhong:
Restores "Add Map Layer" option in the Meta menu

- Menu text changed to "Add raster tile layer"
- Adding the raster layer sets the map style to a new style with only the raster layer and the geojson data/draw layers
- Adding the raster layer appends a button to the style selection UI
- User can switch between the custom style and the core styles until reload
- If user adds another raster layer when one has already been added, it will overwrite the existing one

### DIFF
--- a/src/lib/meta.js
+++ b/src/lib/meta.js
@@ -1,91 +1,120 @@
+const { map } = require('d3');
 var escape = require('escape-html'),
-    geojsonRandom = require('geojson-random'),
-    geojsonExtent = require('geojson-extent'),
-    geojsonFlatten = require('geojson-flatten'),
-    polyline = require('polyline'),
-    wkx = require('wkx'),
-    zoomextent = require('../lib/zoomextent');
+  geojsonRandom = require('geojson-random'),
+  geojsonExtent = require('geojson-extent'),
+  geojsonFlatten = require('geojson-flatten'),
+  polyline = require('polyline'),
+  wkx = require('wkx'),
+  zoomextent = require('../lib/zoomextent');
 
 module.exports.adduserlayer = function(context, _url, _name) {
-    var url = escape(_url), name = escape(_name);
-    var layer = L.tileLayer(url, {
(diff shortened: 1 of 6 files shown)


chriswhong:
Use `https://stamen-tiles.a.ssl.fastly.net/watercolor/{z}/{x}/{y}.jpg` to test

src/lib/meta.js
```diff
@@ -1,91 +1,120 @@
...
-    var url = escape(_url), name = escape(_name);
-    var layer = L.tileLayer(url, {
-        maxZoom: context.map.getMaxZoom()
+  var url = escape(_url), name = escape(_name);
+
+  // reset the control if a user-layer was added before
+  d3.select('.user-layer-button').remove();
+
+  function addUserSourceAndLayer() {
+    // if the source and layer aren't present, add them
```
kuanb, line 17: Would a conditional check wrapping this be useful here?

kuanb (approved):
Approved w/ 1 small comment/question (see inline) and an optional suggestion: perhaps share in issue comment screen cap of new functionality for posterity.

src/lib/meta.js
chriswhong: I don't think so, remove will just do nothing if the selection is empty.

chriswhong:
Oct-21-2022 13-47-44

wafisher:
Hi @chriswhong! Is there a reason the input URL has to be escaped? I have a URL like:

```
https://us0.nearmap.com/maps/z={z}&x={x}&y={y}&version=2&nml=Vert&client=wmts_integration
&httpauth=false&apikey=blah
```

and the ampersands are getting escaped per [this](https://github.com/mapbox/geojson.io/pull/715/files#diff-e79c5a0cd50c6b4292b8660ea8111f82c570f9be0a7d97f5828c06dfe1f8ae7aR11) and the URL that's being requested is wrong.

PR #717, opened 2022-10-20
improve hot reload setup

chriswhong:
Improves the hot reloading configuration.

- `Run on save` only watches the directories and files that are processed by `make`, runs `make` when any of them change.
- `Live Server` ignores the same directories and files, but hot reloads on all others.
- Adds a `wait` setting to `Live Server` which prevents multiple sequential reloads as `make` saves multiple files simultaneously.
- Updates the readme with more info on this setup

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,14 +2,15 @@
   "emeraldwalk.runonsave": {
     "commands": [
       {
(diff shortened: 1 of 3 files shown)


kuanb (approved):
lgtm!

PR #726, opened 2022-10-21
Pass writeable down to addMarkers

### LINKED ISSUE #721: Support marker-size and marker-symbol simplestyle properties
Before the latest release, there was the ability to edit marker properties — `marker-color`, `marker-size`, `marker-symbol`, or add a custom attribute.  Now it is broken. The Edit tool which appears upon clicking on a marker is read-only now.

wboykinm:
Should resolve #721 

@chriswhong Lemme know what you think here; it feels dangerous to set a parameter to a default like this, but I couldn't find another way to do it.

### DIFF
--- a/src/ui/map/util.js
+++ b/src/ui/map/util.js
@@ -21,7 +21,7 @@ const addIds = (geojson) => {
   return geojson;
 };
   
-const addMarkers = (geojson, map, context, writable) => {
+const addMarkers = (geojson, context, writable) => {
   markers.forEach((d) => {
     d.remove();
   });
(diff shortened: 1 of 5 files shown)


chriswhong (changes requested):
This isn't the right fix.  We should be passing `writable` down through `geojsonToLayer()` and then into `addMarkers()`.

I'll push a quick fix to this branch

chriswhong:
We were calling `addMarkers()` without passing in `writable`

chriswhong (approved):

src/ui/map/util.js
```diff
@@ -61,15 +61,15 @@ const addMarkers = (geojson, map, context, writable) => {
...
 };
-function geojsonToLayer(geojson, map, context) {
-  if (map.isStyleLoaded()) {
-    map.getSource('map-data').setData(addIds(geojson));
+function geojsonToLayer(geojson, context, writable) {
+  if (context.map.isStyleLoaded()) {
+    context.map.getSource('map-data').setData(addIds(geojson));
   
-    addMarkers(geojson, map, context);
+    addMarkers(geojson, context, writable);
```
wboykinm, line 72: @chriswhong Man, nicely done finding all the occurrences of this.

wboykinm (commented):
GTG

wboykinm:
image

PR #727, opened 2022-10-21
update default view, fix mapDefault() check

### LINKED ISSUE #724: "L[eaflet] is not defined" error when loading JSON via github url
When trying to load an external geojson file via github url, an error occurs and the geojson is not displayed.

**How to reproduce**

Go to https://geojson.io/#id=github:ebrelsford/geojson-examples/blob/master/brooklyn.geojson

The following appears at the top left, the map is centered on the whole north america instead of brooklyn in the example.

image

chriswhong:
Updates the function that checks whether the map has moved from its default position which previously depended on Leaflet.

To test, load the local server with `#id=github:ebrelsford/geojson-examples/blob/master/brooklyn.geojson` appended to the URL.

Note: There is still a race condition that may prevent the markers from showing up in the above example, this is logged in #728 

This PR will get rid of the Leaflet error, but the map data may not display until we address the race condition.

Fixes #724

### DIFF
--- a/src/core/loader.js
+++ b/src/core/loader.js
@@ -28,7 +28,7 @@ module.exports = function(context) {
   }
 
   function mapDefault() {
-    return context.map.getZoom() == 2 || context.map.getCenter().equals(new L.LatLng(20, 0));
+    return context.map.getZoom() == 2 || JSON.stringify(context.map.getCenter()) === JSON.stringify({lng: 20, lat: 2});
   }
 
   function inlineJSON(data) {
--- a/dist/site.js
+++ b/dist/site.js
@@ -29477,7 +29477,7 @@ module.exports = function(context) {
   }
 
   function mapDefault() {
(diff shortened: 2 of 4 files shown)


src/core/loader.js
```diff
@@ -28,7 +28,7 @@ module.exports = function(context) {
   }
 
   function mapDefault() {
-    return context.map.getZoom() == 2 || context.map.getCenter().equals(new L.LatLng(20, 0));
+    return context.map.getZoom() == 2 || JSON.stringify(context.map.getCenter()) === JSON.stringify({lng: 20, lat: 2});
```
wboykinm, line 31: Elegant and simple. Love it.

wboykinm (approved):
Indeed, the feature does not show up on load (though it shows up when I swap baselayers 🤷), but this tackles the Leaflet helper perfectly.

PR #730, opened 2022-10-21
prevent line or polygon popup if marker was clicked

### LINKED ISSUE #710: Clicking overlapping features opens multiple popups
See https://github.com/mapbox/geojson.io/pull/703#issuecomment-1284133744

For clarification, clicking overlapping linestring or polygon layers doesn't trigger this, it only happens when clicking a marker that overlaps with a polygon or linestring.  This is because there is a separate event listener attached to the html marker element.

Event.stopPropagation can probably help.  https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation

chriswhong:
Prevents two popups from appearing when a marker located within a polygon is clicked.

Closes #710

Also adds `drawing` state boolean to prevent popup from showing when a drawn feature's last vertex is within an existing geometry.  (simple example: draw a new point feature within an existing polygon, the polygon's popup will open)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 ## 2022-10-24
-
+* prevent line or polygon popup if marker was clicked (#730)
 * Fix the handling of points in MultiPoint and GeometryCollection geometries (#736)
 
 ## 2022-10-21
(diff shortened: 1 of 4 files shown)


jcritchley (approved):
Looks good, ready to go!

PR #731, opened 2022-10-22
move codemirror to package.json dependencies

chriswhong:
Removes codemirror from `/lib`, adds it as a dependency in `package.json`

The `Makefile` adds the appropriate javascript to `dist/lib.js` and builds the appropriate css into `css/codemirror.css`


package.json
```diff
@@ -11,6 +11,7 @@
     "clean-css": "~2.0.7",
     "clone": "0.1.16",
     "csv2geojson": "3.6.1",
+    "codemirror": "5.65.9",
```
rreusser, line 14: I see codemirror is at 6.0.1 now. I 5.65.9 gets the job done, I 150% support. Is there any pro/con to upgrading?

rreusser (approved):
Everything looks in order to me! Much cleaner.

package.json
chriswhong: Codemirror 6 is a big change, IIRC it doesn't lend itself well to how we are building this site (we are manually pulling js files out of node_modules and I don't think it was as straightforward to work into our browserify build)

rreusser: Sounds good. 💯

PR #733, opened 2022-10-24
check for source instead of isStyleLoaded()

### LINKED ISSUE #728: Race Condition: data update prior to map load
When loading geojson from a URL parameter such as with a github file, it's possible for `context.data` to be set prior to the map load.  

There is a check for `map.isStyleLoaded()` before updating the map data.  If it runs too soon, the user will end up with properly-loaded geojson in the data editor pane, but nothing on the map.

Example: `https://geojson.io/#id=github:ebrelsford/geojson-examples/blob/master/brooklyn.geojson`

chriswhong:
Closes #728 

`geojsonToLayer()` is fired when `context.data` changes, triggering a `map.setData()` and re-rendering of markers. 

It was checking for `map.isStyleLoaded()` which was returning false and preventing the initial setData() from  running.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2022-10-25
+
+* Fix a bug that was checking for map load in an unreliable way before adding data from externally loaded geojson (#733)
+
 ## 2022-10-24
 
 * Fix the handling of points in MultiPoint and GeometryCollection geometries (#736)
--- a/dist/site.js
+++ b/dist/site.js
@@ -41877,9 +41877,9 @@ const addMarkers = (geojson, context, writable) => {
 };
 
 function geojsonToLayer(geojson, context, writable) {
-  if (context.map.isStyleLoaded()) {
-    context.map.getSource('map-data').setData(addIds(geojson));
-  
+  const dataLoaded = context.map.getSource('map-data');
+  if (dataLoaded) {
+    dataLoaded.setData(addIds(geojson));
     addMarkers(geojson, context, writable);
   }
 }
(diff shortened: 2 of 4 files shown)


wboykinm (commented):
@chriswhong This works exactly as advertised - that brooklyn test layer loads right up on the first map render.

One Q though: Am I incorrect in recalling that you had set geojson layers to render _under_ the labels? If so, is this a regression? 

image

I'm having trouble figuring out the mechanism by which this would be happening. But I'm finding

1. If I load the brooklyn polygon per the issue test, it renders _above_ all labels, and any subsequently-drawn features are also above the labels
2. If I begin by drawing a feature, it renders _under_ the labels as before.

chriswhong:
There's no code anywhere that would cause the mapbox-gl-draw layers or the working dataset layers to appear below labels, so I would ask you to look again re: #2.

[I just checked and we've never had `before` happening in and addLayer() since the GL upgrade.](https://github.com/mapbox/geojson.io/blob/f6e8a9805f781ca58ae3ac3d30a4a51a709ffe84/src/ui/map.js#L301-L332)

wboykinm:
Man, I think it must be baked into `mapbox-gl-draw` then, because drawn layers totally appear under labels [crucial caveat] _when in globe view_, not when in Mercator.

image

image

I'm going to peel this off into a different issue; your PR is solid.

wboykinm (approved):
Great catch on the `dataLoaded` condition!

chriswhong:
Very interesting! I see it too.  Confirmed that the layers as listed in `map.getStyle().layers` show the working dataset layers and the gl-draw layers above labels, but the labels are definitely showing on top in globe view.
geojson_io

PR #734, opened 2022-10-24
prevent editor data update when data is unchanged

chriswhong:
The JSON editor is both responding to changes when `context.data` is updated, and updating `context.data` when changes are detected in the editor.  This creates circular logic where the editor tries to set `context.data` to the same value it just received.

1) editor detects change in `context.data`, sets editor value
2) editor picks up this change as an edit, sets `context.data`
3) editor detects change in `context.data` but knows it was initiated by the editor so it ignores it.

Once data is loaded from a URL parameter or file, the same sequence repeats.

This PR interrupts #2 above and will only set `context.data` if the current value in the text editor is different (meaning the user has edited the geojson) 

The bigger reason to fix this inefficiency is because each time `context.data` is updated it dispatches a `change` event, which is picked up by the map and causes `setData()` on the data source AND removes and re-renders all the markers.  The latter can be costly and cause a visual "blip", and should only be called when there is actually new data to render.

### DIFF
--- a/src/panel/json.js
+++ b/src/panel/json.js
@@ -1,3 +1,4 @@
+var _ = require('lodash');
 var validate = require('../lib/validate'),
   zoomextent = require('../lib/zoomextent'),
   saver = require('../ui/saver.js');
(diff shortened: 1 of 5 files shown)


src/panel/json.js
```diff
@@ -54,8 +55,11 @@ module.exports = function(context) {
 
     function changeValidated(err, data, zoom) {
       if (!err) {
-        context.data.set({map: data}, 'json');
-        if (zoom) zoomextent(context);
+        // don't set data unless it has actually changed
+        if (!_.isEqual(data, context.data.get('map'))) {
```
wboykinm, line 59: Great catch!

wboykinm (approved):
LGTM - side note, is this the first use of lodash or is it leveraged elsewhere in the repo?

chriswhong:
This is the first use of lodash directly... I recall needing it but doing a JSON.stringify() comparison last week, so maybe we should clean that up.

Issue #732, opened 2022-10-23
Imported/pasted geojson is not consistently normalized into a FeatureCollection


chriswhong:
In poking around the codebase recently I noticed the dependency `geojson-normalize` which converts standalone features or geometries into a `FeatureCollection`.  

This is applied when opening a file, so the working data in geojson.io will always be a FeatureCollection.  Features in a FeatureCollection map well to rows in a table, so it makes sense to force everything to be a FeatureCollection so the table in geojson.io will work as expected.

However, there are other ways to get data into geojson.io, specifically loading from a hosted geojson file, gist, or github file, or simply pasting data into the editor.  These methods for data entry to not pass the data through `geojson-normalize`, so we end up in a state where the working data is a standalone feature or even just a geometry.  The editor validates it as valid geojson. 

For consistency, we should ensure that before `context.data.set({map: geojson})` is ever called, `geojson` is a FeatureCollection.  This will ensure that we get predictable results when exporting, when rendering markers, and when rendering the data table.

PR #736, opened 2022-10-24
handle markers for Point, MultiPoint, and GeometryCollection geometries

### LINKED ISSUE #722: Point geometry is not being displayed when part of GeometryCollection
When I try to render a GeometryCollection that contains a single Point and 3 Polygons, the Polygons are being displayed correctly but the Point does not.  This worked a couple days ago.  

I have validated the GeoJSON is correct by displaying at geojsonlint. Attached is the sample FeatureCollection containing the GeometryCollection.

[GeometryCollectionWithPoint.zip](https://github.com/mapbox/geojson.io/files/9832906/GeometryCollectionWithPoint.zip)

chriswhong:
Updates `addMarkers()` to properly render markers from each of `Point`, `MultiPoint`, and `GeometryCollection` geometries.

- Assumes geojson is always a FeatureCollection (see #732)
- Properties are passed down from MultiPoints and GeometryCollections into a new Feature with Point geometry which will power the marker.  The properties will show up in the popup, and editing properties in a popup will edit them for the entire MultiPoint or GeometryCollection.

Closes #722

### DIFF
--- a/dist/site.js
+++ b/dist/site.js
@@ -41802,24 +41802,60 @@ const addIds = (geojson) => {
 };
   
 const addMarkers = (geojson, context, writable) => {
+  // remove all existing markers
   markers.forEach((d) => {
     d.remove();
   });
   let pointFeatures = [];
-  let pointFeaturesIndices = [];
-  // add markers
-  if (geojson.features) {
-    geojson.features.forEach((d, i) => {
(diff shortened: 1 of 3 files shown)


sakethpalla (approved):
Changes look good and testing results confirm that Point, MultiPoint and GeometryCollection geometries all work now:

image

PR #748, opened 2022-10-28
add circle drawing mode, add title tooltips for draw buttons

chriswhong:
Oct-28-2022 00-58-25

- Adds circle drawing mode, which uses `turf.circle()` to create a circle-like polygon with 64 vertices.
- To draw a circle, user clicks the center point, then clicks again at the desired radius.
- Adds `title` properties to all draw buttons, indicating their functionality when hovered.


wboykinm:
I'm a little tempted to disable the circle editor for mercator view.

Both because of the inevitable comments about geodetic accuracy and ovals . . .

image

. . . as well as for the downright-weird things that happen with this tool in certain cases (in this case if you extend the radius too far beyond a pole):

image

chriswhong:
You're right, this needs more thinking, `turf.circle()` is not the right approach.  

Maybe this technique can be used https://stackoverflow.com/questions/50594891/d3-geojson-geocircle-ellipse-equivalent

chriswhong:
I misunderstood the concern and thought it was because `turf.circle` was not generating proper geographic circles.  I think it's fine for it to act kooky near the poles.  Maybe we can add a note in the help column warning about this.

I think it's worth keeping around in Mercator mode as it illustrates the distortion pretty well!
geojson_io

wboykinm (approved):
This seems very much GTG; sorry for the misdirection.

nashitshayan:
hey @chriswhong , I've been trying to implement this circle draw functionality for a project and I've got almost everything working but I can't figure/find out how you managed to show the measurements as the radius is changed. Would really appreciate if you can point me in the right direction here. Thanks!

Issue #740, opened 2022-10-24
Working dataset layers not visible after switching to OSM style


ChrisFleet:
This is a specific problem when a raster tile layer has been added, using 'Meta > Add raster tile layer'. An opened GeoJSON file which displays fine on top of the added raster tile layer, and displays fine on all the other layers, ie. 'Streets | Satellite Streets...' etc. disappears when the OSM layer is selected.

chriswhong:
I was able to reproduce this, it looks like `style.load` event is not firing but only on certain transitions (like from the OSM style to the user-defined raster style)

We listen for `style.load` after the user switches base styles, then we re-add the source and layers for the working geojson dataset.  https://github.com/mapbox/geojson.io/blob/gh-pages/src/ui/map/index.js#L205-L254

Instead of listening for the event we could probably just trigger this manually when the user clicks the button for one of the styles.

chriswhong:
Fixed in #751

ChrisFleet:
Many thanks @chriswhong - I've just tested this now, and you're quite right, its all sorted. Very grateful to you.

PR #751, opened 2022-10-29
fix style load check

chriswhong:
Instead of listening for `style.load`, which is not part of the public api, we add `mapStyleLoaded` as a boolean to `context.data`.  When the user clicks a different style, `mapStyleLoaded` is set to `true`.  On the map `idle` event, if `mapStyleLoaded` is true and the map does not already contain the working dataset sources and layers, we add them and set `mapStyleLoaded` to false.  

This method also works when toggling between a user-added raster layer and any of the existing base styles (fixing #740)

### DIFF
--- a/dist/site.js
+++ b/dist/site.js
@@ -57953,7 +57953,8 @@ function _getData() {
     dirty: false,
     source: null,
     meta: null,
-    type: 'local'
+    type: 'local',
+    mapStyleLoaded: false
   };
 }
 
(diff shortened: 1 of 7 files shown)


dist/site.js
```diff
@@ -57897,7 +57902,7 @@ module.exports = function fileBar(context) {
             alt: 'Add a custom tile layer',
             action: function () {
               var layerURL = prompt(
-                'Layer URL\ne.g. http://stamen-tiles-b.a.ssl.fastly.net/watercolor/{z}/{x}/{y}.jpg'
+                'Layer URL\ne.g. https://stamen-tiles-b.a.ssl.fastly.net/watercolor/{z}/{x}/{y}.jpg'
```
wboykinm, line 57905: Nice catch ✋

wboykinm (approved):
@chriswhong Works perfectly. Nice work rearranging the states here.

PR #752, opened 2022-10-29
Persist projection and style

### LINKED ISSUE #719: Store Globe/Mercator preference in cookie
I much prefer working with a flat map rather than a globe. (Sorry, everyone at Mapbox who worked very hard to make globes work :))

Currently geojson.io reverts back to a globe every time you open it, even when going back to a previous map.

Could the preference please be persisted?

(I'd also like to humbly suggest you consider making Mercator the default. I don't think globe is the right choice for editing datasets at global scale. Particularly since you can't see the whole world at once, and hence, potentially not the whole dataset at once.)

chriswhong:
- Adds `projection` and `style` to localstorage, so the user's selections for both options will persist between reloads.

- Adds `constants.js` with defaults for both values which are used in more than one place (both the UI and map init need to check for the existence of these values in localstorage or use the default)

Closes #719

### DIFF
--- a/src/ui/map/index.js
+++ b/src/ui/map/index.js
@@ -43,7 +43,7 @@ module.exports = function (context, readonly) {
       style: 'mapbox://styles/mapbox/streets-v11',
       center: [20, 0],
       zoom: 2,
-      projection: 'globe',
+      projection: context.storage.get('projection') || 'globe',
       hash: 'map',
     });
 
--- a/src/ui/map/index.js
+++ b/src/ui/map/index.js
@@ -5,6 +5,8 @@ const DrawRectangle = require('../draw/rectangle');
 const ExtendDrawBar = require('../draw/extend_draw_bar');
 const { EditControl, SaveCancelControl, TrashControl } = require('./controls');
 const { addIds, addMarkers, geojsonToLayer, bindPopup } = require('./util');
+const styles = require('./styles');
+const { DEFAULT_STYLE, DEFAULT_PROJECTION } = require('../../constants');
 
 let writable = false;
 let drawing = false;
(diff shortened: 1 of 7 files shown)


src/ui/map/index.js
```diff
@@ -38,12 +40,17 @@ module.exports = function (context, readonly) {
   function map(selection) {
     mapboxgl.accessToken = 'pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpejY4NXFhYTA2bTMyeW44ZG0ybXBkMHkifQ.gUGbDOPUN1v1fTs5SeOR4A';
 
+    // 
+    const projection = context.storage.get('projection') || DEFAULT_PROJECTION;
+    const activeStyle = context.storage.get('style') || DEFAULT_STYLE;
```
wboykinm, line 45: This is really cool to see in action.

src/ui/map/styles.js
```diff
@@ -0,0 +1,48 @@
+module.exports = [
```
wboykinm, line 1: Great idea to pull this out into its own concern, though I think we should discourage additional default styles; we're approaching the maximum number for the UI (and we're over it for the mobile UI)

wboykinm (approved):
Works like a charm 🚢

src/ui/map/index.js
chriswhong: It just works. 🔥

src/ui/map/styles.js
chriswhong: +1

src/ui/map/index.js
wboykinm: 🪄

PR #753, opened 2022-10-30
upgrade fontawesome, add copy button to json editor

### LINKED ISSUE #729: Add a copy button to the data editor
Maybe in the empty space in the tab bar?
geojson_io

chriswhong:
- Upgrades fontawesome to v6 (updates classes for existing icons to use `fa-` prefix instead of `icon-` prefix)
- Updates Makefile to include a command for pulling the necessary files from fontawesome in node_modules and writing them to `/css/fontawesome`

- Adds a copy to clipboard button to the json editor, with success mode (check mark) and tooltip.
Fullscreen_10_30_22__12_04_AM

Closes #729


index.html
```diff
@@ -8,7 +8,9 @@
   <link href='https://api.tiles.mapbox.com/mapbox-gl-js/v2.9.2/mapbox-gl.css' rel='stylesheet' />
   <link rel="stylesheet"
     href="https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v5.0.0/mapbox-gl-geocoder.css" type="text/css">
-  <link rel="stylesheet" href="css/font-awesome/css/font-awesome.min.css" />
+  <link rel="stylesheet" href="css/fontawesome/css/fontawesome.min.css" />
+  <link rel="stylesheet" href="css/fontawesome/css/brands.min.css" />
```
wboykinm, line 12: @chriswhong For some reason I'm still getting a 404 on the `brands` file even after running `make` successfully. It's not a blocker here though - the icons in use still work great.

Makefile
```diff
@@ -71,6 +72,11 @@ css/codemirror.css:
 	cat node_modules/codemirror/lib/codemirror.css \
 	node_modules/codemirror/addon/fold/foldgutter.css > ./css/codemirror.css
 
+css/fontawesome.css:
+	cat node_modules/@fortawesome/fontawesome-free/css/fontawesome.min.css > ./css/fontawesome/css/fontawesome.min.css
+	cat node_modules/@fortawesome/fontawesome-free/css/solid.min.css > ./css/fontawesome/css/solid.min.css
+	cp -R node_modules/@fortawesome/fontawesome-free/webfonts ./css/fontawesome
```
wboykinm, line 78: I should know this by now, but will the makefile always be run on a gh action or via some similar publishing step?

wboykinm (approved):
@chriswhong GTG here; no blocking Qs. That copy function is 🔥

Makefile
chriswhong: The only place it is automated is in the settings for the `runonsave` vs code extension.  The built files must be committed to show up on the live site.

chriswhong: Actually now that you have flagged this I am seeing that this line should not be included under `all` in the makefile, as these fontawesome dependencies are committed once and will never change until we need to upgrade the fontawesome package or bring in other parts of it.  `all` is what runs when we use the command `make` and should only include the steps that depend on source code changes.  

I amended a change just now to remove them from the `all` command.

PR #754, opened 2022-10-30
fix sidebar tab alignment, fix gl control margin in fullscreen mode

chriswhong:
Fixes a misalignment of the sidebar tabs caused by the widened codemirror gutter. geojson_io

Bumps the mapboxgl top right controls down when in fullscreen mode so there is no overlap with the sidebar header.

geojson_io

### DIFF
--- a/css/site.css
+++ b/css/site.css
@@ -1,572 +1,588 @@
 html {
-    height: 100%;
-  }
-  .fillD {
-    background: #ebebeb;
-  }
-  
-  .inline {
-    display: inline-block;
-  }
-  
-  .pad0x {
-    padding: 2px 5px;
-  }
-  
-  /* Utility classes */
-  .map {
-    position: absolute;
-    top: 0;
-    bottom: 0;
-    left: 0;
-    right: 40%;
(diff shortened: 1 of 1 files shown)


css/site.css
```diff
@@ -1,572 +1,588 @@
...
+
+button.delete-invert:hover {
+  color: #e25353;
+}
+
+.top .collapse-button {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 40px;
```
rreusser, line 171: Looks good!

css/site.css
```diff
@@ -1,572 +1,588 @@
...
+  color: #222;
+}
+
+.top .info:hover {
+  background: #ccc;
+}
+
+.top .buttons {
+  margin-left: 40px;
```
rreusser, line 105: 💯

rreusser (approved):
A tiny request for ease of review would be to try to separate PRs with pure-noise formatting changes from PRs with actual content so that the changes can be located, but everything here looks 💯 to me 👍

PR #761, opened 2022-11-05
install mapboxgl, mapbox-gl-draw, mapbox-gl-geocoder, and turf from n…

chriswhong:
- Instead of loading `mapboxgl`, `mapbox-gl-draw`, `mapbox-gl-geocoder` and `turf` from cdn links, include them in `package.json` and require them in the code so they are bundled by browserify.

- Instead of loading all of turf, include only the modules we are using (`area`, `bbox`, `circle`, `length`)

- Updates `Makefile` to bundle the CSS for `mapboxgl`, `mapbox-gl-draw`, `mapbox-gl-geocoder` into one CSS file `mapbox-gl-bundle.css`


package.json
```diff
@@ -7,7 +7,13 @@
     "@fortawesome/fontawesome-free": "^6.2.0",
     "@mapbox/gist-map-browser": "0.2.1",
     "@mapbox/github-file-browser": "0.6.1",
+    "@mapbox/mapbox-gl-draw": "^1.3.0",
+    "@mapbox/mapbox-gl-geocoder": "^5.0.1",
     "@popperjs/core": "^2.11.6",
+    "@turf/area": "^6.5.0",
+    "@turf/bbox": "^6.5.0",
+    "@turf/circle": "^6.5.0",
+    "@turf/length": "^6.5.0",
```
wboykinm, line 16: Great call here. I think this will be a noticeable savings.

index.html
```diff
@@ -5,16 +5,13 @@
...
+  <link href='css/mapbox-gl-bundle.css' rel='stylesheet' />
   <link rel="stylesheet"
-    href="https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v5.0.0/mapbox-gl-geocoder.css" type="text/css">
   <link rel="stylesheet" href="css/fontawesome/css/fontawesome.min.css" />
   <link rel="stylesheet" href="css/fontawesome/css/solid.min.css" />
   <link rel="preconnect" href="https://fonts.googleapis.com" />
   <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
   <link href="https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;700;800&display=swap" rel="stylesheet" />
-  <link rel='stylesheet' href='https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-draw/v1.3.0/mapbox-gl-draw.css'
-    type='text/css' />
```
wboykinm, line 17: Three fewer calls! ✋

package.json
wboykinm: Where/when was the old turf metapackage removed from `package.json`? Or was it always bundled elsewhere?

wboykinm (approved):
Great streamlining, and all the turf functions are still firing on all cylinders for me. 
image

package.json
chriswhong: Turf was being brought in via a CDN link https://github.com/mapbox/geojson.io/pull/761/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051L32

wboykinm: Ahhhhhh.

PR #762, opened 2022-11-05
require codemirror from json.js

chriswhong:
This PR tidys up the `Makefile` a bit.  Codemirror was being bundled along with js files in `/lib`, but it makes more sense to just require it in the file that uses it so it will get picked up by browserify.  (This approach is more consistent with the way we are including other packages from node_modules)


wboykinm (approved):
Makes things satisfyingly compact, and `make` works just fine for me. 🚢

PR #764, opened 2022-11-06
Update linting and formatting

chriswhong:
Introduces eslint and prettier with config similar to other Mapbox frontend repos.

- Updates eslint to use `@mapbox/eslint-config-mapbox/base`, `eslint-config-prettier`, and `eslint-plugin-prettier`. 
- Adds prettier
- Adds `@vercel/git-hooks` and a `git-pre-commit` script to run both prettier and eslint on js on staged files before commit.

Brings all js files in `/src` into compliance with the new linting and formatting rules.

As we don't have tests in place and this touches a lot of files, please take some additional time to manually test things and make sure we haven't introduced regressions.


package.json
```diff
@@ -55,16 +55,29 @@
...
+    "eslint-config-prettier": "^8.5.0",
+    "eslint-plugin-prettier": "^4.2.1",
+    "lint-staged": "^13.0.3",
+    "prettier": "^2.7.1",
     "tailwindcss": "^3.1.8",
     "tape": "~2.12.3"
   },
   "scripts": {
+    "lint": "eslint src/**/*",
+    "format": "prettier src/**/*",
```
rreusser, line 72: This just prints a lot of formatted code for me but doesn't actually write. I think we want [the write option](https://prettier.io/docs/en/cli.html) enabled. This also does not pick up changes in `src/`, for example `src/ui.js`. I think we want `**` instead of `**/*`.

```suggestion
    "format": "prettier --write src/**",
```

rreusser (changes requested):
This looks good to me. I tried to give it a pretty thorough check. The only things I noticed were a couple nits in the scripts, which cause it not to run on all files when invoked imperatively. The scripts probably aren't all that useful anyway since `lint-staged` does work correctly. I think we should either remove them or fix them and keep them (my vote is for the latter since I think it's useful to be able to invoke directly.)

.eslintrc
```diff
@@ -1,27 +1,14 @@
 {
+    "extends": [
+        "@mapbox/eslint-config-mapbox/base",
+        "prettier"
```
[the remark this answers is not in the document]
chriswhong: This setup was borrowed from another Mapbox frontend codebase.  Hoping they do play well together!

package.json
chriswhong: fixed, nice catch

rreusser (approved):
Looks great to me!

PR #765, opened 2022-11-06
add playwright tests for basic drawing

chriswhong:
Adds tests for the following using playwright
- draw a marker, get expected geojson 
- draw a linestring, get expected length measurement, get expected geojson
- draw a polygon, get expected geojson

- Updates `README.md` to include a section about how to run the tests.

Recommend keeping this as a draft until we have attacked more of the checklist and it becomes a more useful test suite.

See checklist in https://github.com/mapbox/geojson.io/issues/772


carlosrziegler:
Nice! <good to learning about tests with Mapbox :)

PR #718, opened 2022-10-20
restore simplestyle properties

chriswhong:
Restores simplestyle-spec styling based on the following properties:

points
- `marker-color`

linestrings
- `stroke`
- `stroke-width`
- `stroke-opacity` 

polygons
- `stroke`
- `stroke-width`
- `stroke-opacity` 
- `fill`
- `fill-opacity`

Closes #712

Issue #498, opened 2016-02-23
Style properties are always saved, even when "not shown"


stevage:
1. Create a point
2. Edit properties
3. Uncheck "Show style properties"
4. Add some other property
5. Save
6. The generated code includes the default styles.

Basically, if the property editor window gets opened, default styles get saved into the file.

tomscholz:
I don't see the problem.

stevage:
The problem is the editor is adding a bunch of SimpleStyle properties which the user didn't ask for, and can't avoid. That would be (maybe) fine if you're generating a file for some application which interprets SimpleStyle, but in most cases, it's just unwanted noise. 

I very often use geojson.io to create sample data, especially for Mapbox-GL-JS, and I don't ever want these properties. 

It makes using the property editor unusable IMHO.

Here's an example of the code generated through the above process:

```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {
        "marker-color": "#7e7e7e",
        "marker-size": "medium",
        "marker-symbol": "",
        "test": "foo"
      },
      "geometry": {
        "type": "Point",
        "coordinates": [
          154.3359375,
          66.08936427047088
        ]
      }
    }
  ]
}
```

chriswhong:
style properties were removed in the recent UI refresh, others are requesting them to be restored, see https://github.com/mapbox/geojson.io/issues/712

PR #766, opened 2022-11-07
restore simplestyle ui in popups, handle marker-size style property

### LINKED ISSUE #756: Restore simplestyle properties UI
Prior to the GL upgrade, feature editing popups included "style properties" which would populate the properties table with simplestyle-spec values.

Support for rendering most of simplestyle (except for `marker-size` and `marker-symbol`) was restored, so we may as well bring back the color pickers and dropdowns, etc, with some minor changes:

- See #498, the style properties were being saved even if the user checked the "show style properties" checkbox.  This means anyone who edited properties via the popups got style properties whether they wanted them or not.  We should add a button `Add Style Properties` to the popup instead of the show/hide checkbox, so it's more "opt-in"

chriswhong:
Handles the `marker-size` simplestyle-spec property.   (this was not included in #718 for some reason) `marker-icon` is the only remaining unsupported simplestyle property.

Restores simplestyle property creation/editing in the popup UI (Closes #756) with the following changes:

- Instead of a checkbox, there is a text button `Add Simplestyle Properties`.  When clicked, it appends the style properties for the current feature type to the properties table.  These will not take effect until the user clicks the save button.   (the previous implementation always included the style rows, but hid the rows, so they would be added on save whether the user wanted them or not. #498)

- Adds some logic to determine whether to show the `Add Simplestyle Properties` button (do not show it if all of the properties for a given geometry type already exist in the feature's properties.

### DIFF
--- a/css/marker.css
+++ b/css/marker.css
@@ -101,11 +101,11 @@ datalist {
     display: block;
 }
 
-.add-row-button {
+.add-row-button, .add-simplestyle-properties-button {
     color: #2980b9;
 }
 
-.add-row-button:hover {
+.add-row-button:hover, .add-simplestyle-properties-button:hover {
     cursor: pointer;
     color: #199CF4;
 }
--- a/dist/site.js
+++ b/dist/site.js
@@ -59738,9 +59738,11 @@ module.exports = function (context) {
 
     sel.selectAll('.cancel').on('click', clickClose);
 
-    sel.selectAll('.save').on('click', saveFeature);
+    sel.selectAll('form').on('submit', saveFeature);
 
     sel.selectAll('.add').on('click', addRow);
+
+    sel.selectAll('.add-simplestyle-properties-button').on('click', addSimplestyleProperties);
 
     sel.selectAll('.delete-invert').on('click', removeFeature);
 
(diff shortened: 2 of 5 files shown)


jcritchley:
Adding simplestyle properties works well.

For markers I'm only getting one option from the drop down for marker-size, matching the value that is being used. From the code seems should be seeing 3 options small, medium and large. I can manually edit to small or large but still just get one option.

image
image

For data that can have opacity values, would be good to limit to max value 1. Anything over is accepted but turns line or fill white.

image

chriswhong:
@jcritchley The first issue is related to how Chrome (all browsers?) handles the html `datalist` element.  `datalist` lets you have a dropdown-like list of options for a text input where the user can also enter any text they want.  However, if the input has a value in it, the dropdown only shows that value.  ¯\_(ツ)_/¯

We should probably change this input to a `select` just for this property, as there are only three valid values and there is no need for freeform text input.

chriswhong:
@jcritchley good catch on the `max` for opacity inputs.  These constraints are set in the html but were not being enforced because there's not actually a `form` in the popups.  I pushed up a change that introduces the form and prevents the user from saving if the value is outside the min and max.
Fullscreen_11_9_22__12_51_PM

jcritchley (approved):
Nice work, tested and all looks good 🥳

Issue #687, opened 2021-08-18
Allow editing on mobile


panchicore:
Awesome utility, we are using it as part of a data collection for incidents that involve markers, lines, and areas, we just need to copy/paste the JSON to make the backend work. But when we want to use it on android phones, even activating the "request desktop version", it won't allow selecting anything on the map, when the marker icon is selected, it displays the "cancel" tooltip but it won't be possible to pin it on the map to generate the geojson text.

chriswhong:
This should be possible after #779 is in.

chriswhong:
Closed as of #779

PR #779, opened 2022-11-11
Mobile Responsive Layout and CSS updates

### LINKED ISSUE #763: Sidebar tabs layout bug
geojson_io

The width of the "gutter" in codemirror gets wider as the numbers it needs to display get bigger.  The tabs have a fixed margin-left that doesn't move when the gutter gets bigger.

### LINKED ISSUE #770: Inefficiency in delegate.js
`delegate.js` requires mapboxgl only to check `mapboxgl.supported()` and redirect the browser to `/unsupported.html`.

Since delegate.js is not part of the main build (site.js), the browser ends up downloading mapboxgl twice.  

I'm not sure why `delegate.js` needs to be its own file and why we can't just do these redirects in the main js bundle.  

We may also want to think bigger about whether `mobile.html` and `unsupported.html` are even necessary (can we make the main app more mobile-friendly/ show an unsupported message in place of the map)

### LINKED ISSUE #775: Remove delegate.js
https://twitter.com/rtwlz/status/1590763551295410178

the script tag that includes delegate.js was removed in this hotfix to get the site working for the chromebook user above. This issue is for follow-on task of cleaning up/removing `mobile.html` and its associated scripts in favor of making the main site mobile responsive.

`delegate.js` uses `bower` to check which device the user is on.  It checks the user-agent for `type` of device and redirects non-desktops to `mobile.html`, a read-only version of the app (which is useless if you aren't loading an external data source)

This setup is pretty clumsy and we should just remove `delegate.js` and `mobile.html` altogether in favor of making the main site responsive.  This will also eliminate the build of `mobile.site.js`, further simplifying things.

I think the assumption here is that nobody is editing data or working with files on a mobile device.  I'm sure there are exceptions to the rule, but we can have some sensible defaults, but I think we should optimize for what happens when someone clicks a geojson.io link with a url parameter that loads external data.  This seems like the original intent of the mobile.html page (to display data but not allow drawing and editing) and we can just do this by toggling UI off in the main site based on either the viewport width or if the device is a touchscreen.

chriswhong:
- Removes `delegate.js` and `site.mobile.js`, and their associated commands in `Makefile`. 
- Instead of redirecting to `mobile.html` for some browsers, the main site at `index.html` is now mobile responsive.
- Adds a modified flex layout for the main content (map + sidebar) to facilitate mobile responsiveness, also adding a new control and css transition for collapsing the sidebar.
- Updates the layout of the three sidebar tabs to fill the entire sidebar, adds a small border below to visually separate the tabs from the content.

Nov-11-2022 11-41-44
geojson_io

(Closes #775)
(Closes #770 as no longer relevant)
(Closes #763 as no longer relevant)


chriswhong:
TODO: make sure touch actions work to open the popups, #687 shows users who want to edit data (not necessarily draw) on mobile and I think touch events work for most things but not the popups.

wboykinm:
After running `make`, I'm getting this on page load:

```
Uncaught TypeError: Cannot destructure property 'style' of 'styles.find(...)' as it is undefined.
    at Object.map (site.js:79724:13)
    at Array.render (site.js:78094:13)
    at d3_selectionPrototype.call (lib.js:932:12)
    at 217../core/data (site.js:76806:25)
    at o (site.js:1:265)
    at r (site.js:1:431)
    at site.js:1:460
```

. . . maybe just a pathology of my local setup?

chriswhong:
Try blowing away node_modules and `npm install` again?  Or maybe clear localstorage as the .find() referenced above is trying to locate a style based on what's in locatstorage.

wboykinm:
@chriswhong This is looking pretty great! 

image

Three notes:

1. The `sidebar handle` doesn't appear to be visible in portrait layout.
2. I'm getting `Unable to preventDefault inside passive event listener invocation.` when I try to use the circle tool, though the others work fine.
3. I'm not sure how to test the distinction between a click and a tap, since I'm just using the built-in chrome device emulator.

chriswhong:
> The sidebar handle doesn't appear to be visible in portrait layout.

This is by design, you cannot dismiss the sidebar in mobile view.  (in fact even if it is dismissed in desktop view it will still be visible if you resize your browser to mobile view) We can add this later if there is demand, but think we shouldn't crowd the mobile UI for now.

> I'm getting Unable to preventDefault inside passive event listener invocation. when I try to use the circle tool, though the others work fine.

I looked into this a bit and I think I landed on it being something that needs to be sorted out in `mapbox-gl-draw`.  This warrants some more research and we should squash it before merging this PR.

> I'm not sure how to test the distinction between a click and a tap, since I'm just using the built-in chrome device emulator.

To test clicks in mobile view, just resize your browser to be skinny... to test taps, use the chrome device emulator.

chriswhong:
@wboykinm The error `Unable to preventDefault inside passive event listener invocation` is a known issue with `mapbox-gl-draw` on mobile devices and was not introduced by these changes.  I don't think it's actually causing any problems so I guess we will have to live with it for now.

https://github.com/mapbox/mapbox-gl-draw/issues/1019

wboykinm (approved):
@chriswhong Taps and clicks both work great. If we're good with the basic gl.js error, then this should be GTG.

Issue #784, opened 2022-11-16
Conditionally pass validation errors on winding order


kuanb:
Suggestion:
Validation errors on simple polygons related to violation of the winding order rules can be confusing/frustrating when using geojson.io to quickly visualize something. Also, winding order issues are already handled during render by `geojson-vt` apparently.  It seems that, when validation is performed, we can simply allow instances where the winding order is wrong to be allowed to pass as they will still render fine.

Related conversation w/ @dnomadb: https://twitter.com/DnomadB/status/1592712021866610688

Related lines of code: 
https://github.com/mapbox/geojson.io/blob/d10be259a24bd3074e8695e481f94ca777b5bcb3/src/lib/validate.js#L8-L22

Details: Looks like there would just be an additional check to ensure right hand rule is followed. Since we updated `geojsonhint` it looks like there's an option to simply turn off this check as well, per this PR: https://github.com/mapbox/geojsonhint/pull/87

The pattern suggested looks like:
```js
geojsonhint.hint(geojsonWithIncorrectWindingOrder, {
  ignoreRightHandRule: true
});

**Comment by chriswhong** [comment]
Looks like the [geojson spec](https://www.rfc-editor.org/rfc/rfc7946#section-3.1.6) has some guidance here:

```
Note: the [[GJ2008](https://www.rfc-editor.org/rfc/rfc7946#ref-GJ2008)] specification did not discuss linear ring winding
   order.  For backwards compatibility, parsers SHOULD NOT reject
   Polygons that do not follow the right-hand rule.
```

Also, `geojsonhint` includes `level: message` in the objects it returns when they are meant to be warnings, not errors.  (rhr shows up this way)  We aren't checking for this at all.  

I'll take a look at what it would take to show this as a warning without rejecting the geojson as invalid.

chriswhong:
Opened #786 which properly handles "errors" vs "warnings" returned by `geojsonhint`.  Geojson with polygons that violate the right-hand rule will show a warning in the code editor but will not be rejected.

PR #785, opened 2022-11-16
Ignore right hand rule flag

kuanb:
Resolves https://github.com/mapbox/geojson.io/issues/784

@chriswhong if you can check this and make sure that it's safe/all that needs to be done. Looks like there's no hooks for tests automatically - I just made the change in browser on GH.


chriswhong:
Superseded by https://github.com/mapbox/geojson.io/pull/786. (Accept geojson that does not follow the right hand rule, but show a warning in the code editor)

PR #786, opened 2022-11-16
show geojsonhint warnings without rejecting geojson

### LINKED ISSUE #784: Conditionally pass validation errors on winding order

chriswhong:
Status Quo:

geojson in the code editor is validated using `geojsonhint` which returns an array of error objects.  If this array is not empty, the geojson is considered invalid and is not rendered/updated, and the errors are highlighted in red in the codemirror gutter.

In this PR:

- Checks the array returned from geojsonhint for objects that do not contain a `level` property.  These are "rejectable" errors and if any of these exist, reject the geojson and highlight the errors in the same way as the status quo.
- When geojson does not yield any rejectable errors from geojsonhint, accept it, but also highlight any warnings in yellow.

Example: Feature missing `properties` key yields an error, polygon that does not follow the right-hand-rule yields a warning.

Cursor_and_geojson_io___powered_by_Mapbox

Closes #784

### DIFF
--- a/css/site.css
+++ b/css/site.css
@@ -195,8 +195,16 @@ button.delete-invert:hover {
   -webkit-transition: all 100ms;
 }
 
+.error-marker.warning {
+  background: #eee0aa;
+}
+
 .error-marker:hover {
   background: #d68;
+}
+
+.error-marker.warning:hover {
+  background: #eee0aa;
 }
 
 .error-marker:before {
--- a/dist/site.js
+++ b/dist/site.js
@@ -77368,6 +77368,12 @@ module.exports = function (callback) {
     const err = geojsonhint.hint(editor.getValue());
     editor.clearGutter('error');
 
+    // check err for objects that don't have `level` properties
+    // if any exist, reject the geojson
+    const rejectableErrors = err.filter(
+      (d) => !Object.prototype.hasOwnProperty.call(d, 'level')
+    );
+
     if (err instanceof Error) {
       handleError(err.message);
       return callback({
@@ -77375,7 +77381,7 @@ module.exports = function (callback) {
         title: 'invalid JSON',
         message: 'invalid JSON'
       });
-    } else if (err.length) {
+    } else if (rejectableErrors.length) {
       handleErrors(err);
       return callback({
         class: 'icon-circle-blank',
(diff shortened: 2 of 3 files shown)


dnomadb:
🙏 thank you @chriswhong !

kuanb (approved):
Clean diff! One question for myself, but otherwise approved. Thanks!

chriswhong:
Sample data which will trigger a right-hand-rule warning, for posterity:

```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {
        "id": "01",
        "name": "spara",
        "cap": [
          "01",
          "02",
          "03",
          "04",
          "05",
          "06",
          "07",
          "08",
          "09",
          "10"
        ]
      },
      "geometry": {
        "coordinates": [
          [
            [
              [
                10.054115864565517,
                42.26090627280195
              ],
              [
                10.054115864565517,
                41.36139362268415
              ],
              [
                10.936127216436944,
                41.36139362268415
              ],
              [
                10.936127216436944,
                42.26090627280195
              ],
              [
                10.054115864565517,
                42.26090627280195
              ]
            ]
          ],
          [
            [
              [
                11.732750376967573,
                42.07091206517259
              ],
              [
                12.746113089809938,
                42.014470439676046
              ],
              [
                12.594475367391851,
                41.496275845578
              ],
              [
                11.843562558734305,
                41.49592342463356
              ],
              [
                11.732750376967573,
                42.07091206517259
              ]
            ]
          ],
          [
            [
              [
                10.39092142845297,
                40.935511886112124
              ],
              [
                10.39092142845297,
                40.13363469371467
              ],
              [
                11.155195061855636,
                40.13363469371467
              ],
              [
                11.155195061855636,
                40.935511886112124
              ],
              [
                10.39092142845297,
                40.935511886112124
              ]
            ]
          ]
        ],
        "type": "MultiPolygon"
      }
    }
  ]
}
```

PR #787, opened 2022-11-16
normalize pasted data into FeatureCollection

### LINKED ISSUE #732: Imported/pasted geojson is not consistently normalized into a FeatureCollection

chriswhong:
- Adds a normalization step to the validation of data from the code editor.  This ensures that the the working dataset is **always** a FeatureCollection.  Prior to this PR, data imported via the 'Open' menu button or by drag and drop is normalized into a FeatureCollection, but data loaded via pasting geojson into the code editor or by loading from a url or gist is not.  
- Refactors the code for rendering markers to always expect a FeatureCollection.  
- Updates `help` to include a statement about the working dataset always being a FeatureCollection
- Minor change to the description of what geojson.io is, to match the description in the github repo

Examples of inconsistencies prior to this PR:
- If the user pastes/loads a Feature or Geometry, it will appear on the map but there will be nothing in the table view
- If the user pastes/loads  a Feature or Geometry and tries to export to shapefile, an error shows up in the console and no file is created
- If the user pastes/loads  a Feature or Geometry and then tries to create a new Feature with the drawing tools, an error occurs 

Needs thorough manual testing to make sure we haven't caused any regressions.  Things to try:

- Load a lone geometry by pasting in the editor
- Load a lone feature by pasting in the editor
- Export to shapefile or kml after importing the above
- Draw another feature using the drawing tools
- Interact with the table, edit properties in a popup, edit properties via table, edit properties via code editor, make sure edits in one place show up in the others

Closes #732

### DIFF
--- a/data/help.html
+++ b/data/help.html
@@ -1,7 +1,7 @@
 <h2>Help</h2>
 
 <p><strong>geojson.io</strong> is a quick, simple tool for creating,
-    viewing, and sharing maps. geojson.io is named after <a href='http://geojson.org/' target='_blank'>GeoJSON</a>,
+    viewing, and sharing spatial data. geojson.io is named after <a href='http://geojson.org/' target='_blank'>GeoJSON</a>,
     an open source spatial data format, and it supports GeoJSON in all ways - but also
     accepts KML, GPX, CSV, GTFS, TopoJSON, and other formats.</p>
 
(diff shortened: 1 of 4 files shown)


wboykinm (approved):
Wow, yep. Works like magic.

image

. . . and I would say it's _too_ magical except you have that excellent alert popup indicating that the paste has been coerced to a FeatureCollection.

Exports also work correctly, as well as all manner of property fiddling and cross-copying. Nice job!

Issue #610, opened 2018-03-16
"Feature": Error when "properties" is "null"


cweiske:
The GeoJSON specification allows NULL values for the `properties` key:

> A Feature object has a member with the name "properties".  The value of the properties member is an object (any JSON object or a JSON null value).

https://tools.ietf.org/html/rfc7946#section-3.2

However when opening a file with `null` properties, I get an error:

```
Invalid JSON file: TypeError: can't convert null to object
```

Example file that triggers this error: http://p.cweiske.de/556

Happens in Firefox 59.

chriswhong:
The example provided loads without issue now.

cwarden:
Opening a geojson file with null properties is failing again, with "Invalid JSON file: TypeError: Cannot convert undefined or null to object", including the example file linked above, http://p.cweiske.de/556.

PR #790, opened 2022-11-18
handle null properties, refactor layer initialization logic, refactor zoom on first load

### LINKED ISSUE #788: Display point geometry when passed using URL params
hey @chriswhong - thanks so much for giving this repository some love! So happy to see all the new changes 👏 👏 👏 

I use the [geojsonio-cli](https://github.com/mapbox/geojsonio-cli/) a fair amount in my work- it's a little CLI that passes geojson to geojson.io for visualization. I use this most often with point geometries to double check where things are. Here's an example URL generated by the CLI:

http://geojson.io/#data=data:application/json,%7B%22type%22%3A%22Feature%22%2C%22geometry%22%3A%7B%22type%22%3A%22Point%22%2C%22coordinates%22%3A%5B3.7080748129414016%2C51.53710507194907%5D%7D%2C%22properties%22%3Anull%7D

This passes this geojson to geojson.io:
```json
{"type":"Feature","geometry":{"type":"Point","coordinates":[3.7080748129414016,51.53710507194907]},"properties":null}
```

This used to display a point marker on the map, but now seems to just zoom to the point itself. Any chance you could bring this feature back to display a point when passed via the URL? Or is there a new way for me to achieve this functionality?

Here's a screenshot of the current output for the link above. The JSON is visible in the JSON panel, but no marker on the map:

image

Thank you!

chriswhong:
- fixes a bug where adding markers with null properties would crash (closes #788 and #610)

- refactors the initial zoom/fitbounds, which was being called too early and delaying the load of the data (user would see the map animate to the bbox, then the data would appear).  This introduces a 'recovery' variable in `context.data`, if `true`, `geojsonToLayer()` will call `zoomExtent()` and reset `recovery` to `false`

- `zoomExtent()` is refactored to use flyTo() when the working dataset contains a single Point feature (avoids a long `fitBounds` that goes to the highest zoom level because the `bbox` of a point is a point)

- refactors map layer logic to be more efficient:  
`map.on('idle')` only adds the working dataset source (with dummydata) and adds the working dataset layers, then calls `geojsonToLayer()`.  `geojsonToLayer()` is what actually calls `setData()` and `addMarkers()` to get the data updates onto the map.  

The `map.on('idle')` logic needs to happen on first load and when the user changes styles via the UI.

`geojsonToLayer()` is always called after `map.on('idle')` , but is also triggered by data changes (such as drawing a new Feature or editing via the code editor.


wboykinm (approved):
> fixes a bug where adding markers with null properties would crash (closes https://github.com/mapbox/geojson.io/issues/788 and https://github.com/mapbox/geojson.io/issues/610)

This works if the acceptance criteria is to allow `"testkey": null`; `"testkey": ` throws an error (as I'd expect it to).

> refactors the initial zoom/fitbounds, which was being called too early and delaying the load of the data (user would see the map animate to the bbox, then the data would appear). This introduces a 'recovery' variable in context.data, if true, geojsonToLayer() will call zoomExtent() and reset recovery to false

This also works great as advertised. One note is that the data now happily appears before the zoom animation, which for a fleeting milisecond (faster than I can capture in a screenshot, actually) means that any data beyond the horizon from the initial view appears at the edge of the load view. For example, a bunch of points set in Mexico appeared in Ghana on first recovery reload (for a heartbeat), then the zoom animation proceeded and everything looked great.

> zoomExtent() is refactored to use flyTo() when the working dataset contains a single Point feature (avoids a long fitBounds that goes to the highest zoom level because the bbox of a point is a point)

Super smooth. Like butter. 👏 

> refactors map layer logic to be more efficient:
map.on('idle') only adds the working dataset source (with dummydata) and adds the working dataset layers, then calls geojsonToLayer(). geojsonToLayer() is what actually calls setData() and addMarkers() to get the data updates onto the map.

Hard to say how to test this, but no obvious errors thrown on various data updates.

wboykinm:
@chriswhong Review above, but side Q: What's up with all the fonts?

chriswhong:
Good catch, those shouldn't have made it into this PR!  I'll remove them now.

chriswhong:
> One note is that the data now happily appears before the zoom animation, which for a fleeting milisecond (faster than I can capture in a screenshot, actually) means that any data beyond the horizon from the initial view appears at the edge of the load view.

This only happens with markers AFAIK.  I think we can punt on this for now.  I think no matter when you `addMarker()`, if the point is on the other side of the horizon you're going to see it flash.  I am pretty sure this is a mapboxgl issue, but we should investigate further.

PR #792, opened 2022-11-18
update core style urls

chriswhong:
Core styles updates launched today!

https://www.mapbox.com/blog/mapbox-core-styles-redesigns

Edit:  Since some of the core styles include fog, we drop the `setFog()` that happens on each style change.  Instead we use the style's fog settings  for Streets, Satellite-Streets, and Outdoors.  For the other styles we manually set the fog.


jcritchley (approved):
Thanks for the fog property changes. Looks good to go!

PR #794, opened 2022-11-23
fixes #781 by setting a new default color for satellite and dark

### LINKED ISSUE #781: Change default line color for satellite view
The current default is really hard to see close up, since they are about the color and width of the streets. 

As an example, I dragged the Miami Metromover geojson lines over on top of Baltimore. Could you tell where it is?

image

Here is the black-and-white
image

DJFriar:
fixes #781 

I tested a bunch of different colors and this was one of the easier ones to see on the satellite base map, while also still looking good on the dark base map.

For simplicity I just changed the value but it would be easy enough to make this only apply to the Satellite base map and leave the Dark alone if that is the preference.


chriswhong:
Hi, thanks so much for working on this.  I checked out your branch and still think this color doesn't contrast enough with the features in the satellite style, especially when looking at an urban area:

geojson_io___powered_by_Mapbox

At the very least I would suggest we come up with a special color just for satellite that provides good contrast most of the time.  Would you be able to re-work this and try to find a good color, then make sure it's only applied to the satellite-streets style?

Also I haven't checked but we should also make sure the markers are getting these colors when we switch to the different styles.

chriswhong:
geojson_io___powered_by_Mapbox

Disregard my previous comment!  I didn't build the bundle.  We don't have a separate build step, so can you please commit `dist/site.js` to your branch?  (I have some plans for a separate build step in the works, but it's not ready yet)

I don't have strong opinions on this purple and think it solves the immediate problem.  I've opened #796 as a new issue to address the mismatch between marker colors and line/polygon colors.

If you would be interested in attacking #796 in this PR as well that would be great.  If not, no worries, we can merge this and work on the other issue separately.

chriswhong:
If it's not too much to ask, I would suggest we use the purple for satellite and OSM.  So leave `LIGHT_FEATURE_COLOR` as it was before and add a new `PURPLE_FEATURE_COLOR` constant and the logic to apply it to those two styles.

DJFriar:
> If it's not too much to ask, I would suggest we use the purple for satellite and OSM. So leave `LIGHT_FEATURE_COLOR` as it was before and add a new `PURPLE_FEATURE_COLOR` constant and the logic to apply it to those two styles.

This has been completed and checked in. The latest commit also fixes #796.

src/ui/map/util.js
```diff
@@ -77,9 +82,28 @@ const addMarkers = (geojson, context, writable) => {
...
+        break;
+      default:
+        defaultColor = DEFAULT_DARK_FEATURE_COLOR;
+    }
+
+    const color =
+      (d.properties && d.properties['marker-color']) || defaultColor;
     let scale = 1;
 
+    console.log(`color is ${color}`);
```
chriswhong, line 105: Stray console log

src/ui/map/util.js
```diff
@@ -117,6 +141,15 @@ const addMarkers = (geojson, context, writable) => {
         writable
       );
     });
+
+    // Update the dot in the Marker for Dark base map style
```
chriswhong, line 145: Nice touch!

src/constants.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
   DEFAULT_PROJECTION: 'globe',
-  DEFAULT_STYLE: 'Streets'
+  DEFAULT_STYLE: 'Streets',
+  DEFAULT_DARK_FEATURE_COLOR: '#555',
+  DEFAULT_LIGHT_FEATURE_COLOR: '#e8e8e8',
+  DEFAULT_PURPLE_FEATURE_COLOR: '#ff40ff'
```
chriswhong, line 6: Glad we moved these here!

chriswhong (changes requested):

chriswhong:
Looks great and works great, just saw one stray `console.log` that should be removed. 

BTW there is now a #geojson-dot-io channel in the MapboxDevs discord if you are interested. May be helpful if you need clarification on future issues you want to work on.  https://discord.gg/uMpcC5RmJh

DJFriar:
> Looks great and works great, just saw one stray `console.log` that should be removed.

I have checked in an update to remove that console.log (and another I found as well). I changed the color from purple to bright green because when I read the email I saw the first message you sent and thought it was about the purple. Anyway, it is green now, but let me know if you prefer the purple and I can change it real quick.

> BTW there is now a #geojson-dot-io channel in the MapboxDevs discord if you are interested. May be helpful if you need clarification on future issues you want to work on. https://discord.gg/uMpcC5RmJh

I joined there just after I made the earlier commit actually. I am also DJFriar there.

chriswhong (approved):

chriswhong:
The green looks great!  Thanks for this!

PR #795, opened 2022-11-23
Fixes #768 by adding marker popup offsets

### LINKED ISSUE #768: Offset Marker Popups
Nov-07-2022 10-03-56

When clicking a marker, the popup is anchored to the same coordinates as the marker.  Let's add `offset` when binding the popup to the marker to shift it up above the marker.

DJFriar:
Fixes #768 

Added popup offsets to the Marker to account for all 8 anchor points.

### DIFF
--- a/src/ui/map/util.js
+++ b/src/ui/map/util.js
@@ -360,9 +360,21 @@ function bindPopup(e, context, writable) {
       : '') +
     '</form>';
 
+  const popupOffsets = {
+    top: [0, 10],
+    'top-left': [0, 10],
+    'top-right': [0, 10],
+    bottom: [0, -40],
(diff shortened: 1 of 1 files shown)


chriswhong (approved):
Looks great!  I don't think i've ever used offset with this level of detail, I learned something today!

chriswhong:
FYI we generally commit the `/dist` directory so changes can go live on github pages immediately.  I'll merge this and add a commit with the updated bundle.

DJFriar:
> FYI we generally commit the `/dist` directory so changes can go live on github pages immediately. I'll merge this and add a commit with the updated bundle.

Ah, I will do that going forward; thank you for the note. :)

PR #798, opened 2022-11-27
add rollup

chriswhong:
This PR refactors the build tool for geojson.io.  Instead of using `browserify` and a `Makefile`, `rollup.js` handles the building of css and javascript bundles.

**New NPM Scripts**
- `npm start` runs `live-server`, `rollup`, and `npm tailwindcss` - all three commands run concurrently and each is watching certain parts of the codebase for changes.   

This simplifies the development workflow which was dependent on VS code plugins for live reloading, and new developers will be able to get started quickly by installing dependencies and running `npm start`.

- `npm run build` builds minified css and js bundles.  These consolidated/minified bundles dramatically reduce the number and size of files necessary to launch the app in the browser. 
- `npm deploy` runs a shell script that creates a production build and makes an orphan commit on a new branch (including the /dist directory), then force pushes it to the `gh-pages` branch.

**Other Notable Changes**
- This PR's base branch is `main`, which will become the new default branch for this repository.
- `/dist` is no longer committed to the main branch.  `gh-pages` is used for deployment/hosting only and contains only one commit, the build of the latest commit on `main`
- Moving forward, deployments can be run manually on a developer's machine after PRs are merged. We can look into an automated build on new commits to main using github actions.
- `gtfs2geojson` is added to `/lib`.  This was an external npm module but it's dependency `dsv` includes an fs.readFileSync() call that does not play well with rollup.  The version now included in `/lib` replaces `dsv` with its successor package `d3-dsv`.  
- several old/unused polyfills in `/lib` have been removed (aight, base64, blob, etc)


src/lib/gtfs2geojson.js
```diff
@@ -0,0 +1,82 @@
+// vendored gtfs2geojson because the latest published version requires `dsv` which
```
rreusser, line 1: Should the license be inlined here? https://github.com/conveyal/gtfs2json/blob/007dc123acf655a2f06ad3b529facb9d77809434/LICENSE

src/lib/d3-compat.js
```diff
@@ -1,37 +1,36 @@
-(function() {
-
+(function () {
```
rreusser, line 1: Should we drop the d3 license into this directory? https://github.com/d3/d3/blob/main/LICENSE

rreusser (approved):
Looks great to me! I'm not the biggest fan ever of overloading `require()` with other file types, but then it's really nice and clean to handle markdown and css at the bundler level, so I definitely have no qualms with this. Really nice work! :shipit:

rreusser:
Also, regarding `fs.readFileSync`, maybe something like https://www.npmjs.com/package/rollup-plugin-brfs would work? That was always the solution for statically inlining content with browserify (brfs = BRowseify-FS).

chriswhong:
I went down the rabbithole with `rollup-plugin-brfs` and couldn't get it working.  It's not widely used and was last published a few years ago, so I wasn't going to push too hard to get it going.

rreusser:
Sounds good! It's a slam dunk with browserify, but can definitely see that these things can be a huuuuuge rabbit hole when they get a bit outdated. 👍 👍

src/lib/gtfs2geojson.js
chriswhong: Done, but it's from `https://github.com/node-geojson/gtfs2geojson`, not `https://github.com/conveyal/gtfs2json/blob/007dc123acf655a2f06ad3b529facb9d77809434/LICENSE`

src/lib/d3-compat.js
chriswhong: These aren't from d3 to my knowledge, I think they are extensions to d3 written for geojson.io.  These predate our current work and aren't new, they are just moved to a different directory with this PR.

PR #802, opened 2022-11-30
fix css load order

### LINKED ISSUE #801: Regression: polygon icon loads rectangle
Screen Shot 2022-11-29 at 9 43 37 AM

Not clear to me where this would have been introduced if not in the rollup add, though [the CSS shows no substantive change](https://github.com/mapbox/geojson.io/commit/d62731f0426a9a33369ab24c736053674f4859e0#diff-640076e43ea277556b11bc62ec846a740ad8301eb7e730f0b77fe76efaffb0ffR550) (Doublequotes 🤷 ), and the [`polygon.svg` file](https://github.com/mapbox/geojson.io/blob/main/img/polygon.svg) still contains the correct icon.

chriswhong:
Fixes #801 by requiring css in the following order:  tailwind -> mapboxgl -> app css

Tailwind overrides buttons which was causing the mapboxgl mini-attribution control to not display properly, so it is loaded first.  The app css is loaded last, overriding the default button for polygon drawing mapbox-gl-js.

Also moves some changes that were inadvertently made in #800  to `help.html` into `help.md`.  `help.html` is no longer used and is deleted with this PR.

Also also, adds `npm install` to the deploy script.


wboykinm (approved):
PERFECT. 🙇 

image

PR #805, opened 2022-12-08
replace terser with uglify

### LINKED ISSUE #804: Flatten Multi Features not working
Hello everyone,

it seems that the "Meta - Flatten Multi Features" function stopped working.
I was using this regulary over the past year or so to create Polygon type map from MultiPolygon.

But now it just makes the map output blank with:
{
  "type": "FeatureCollection",
  "features": []
}

Tried with two maps now + I also tried a map that I successfully converted in the past - none of them working right now.
Attaching a example file.

Thank you very much for your time and expertise.
[203153.zip](https://github.com/mapbox/geojson.io/files/10180312/203153.zip)

chriswhong:
In troubleshooting #804 I discovered that the minified js bundle (minified by `rollup-plugin-terser`) was causing an error when using `Flatten Multi Features`.  For some reason, the value pulled from `context.data.get('map')` was being mutated immediately after the button is clicked and the action is called.  The `features` array changed to `undefined` and yielded errors with downstream functions that were expecting an array.

This behavior did not occur in development, so it's a direct result of the terser minification.

Rather than go down the rabbithole of figuring out a mysterious mutation, I tried other minification options for rollup and landed on `rollup-plugin-uglify` which seems to solve the problem.

Closes #804

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,6 @@
         "geojson-flatten": "^1.0.4",
         "geojson-random": "0.1.0",
         "geojson2dsv": "0.1.0",
-        "gtfs2geojson": "^2.0.0",
         "lodash": "^4.17.21",
         "numeral": "^2.0.6",
         "osmtogeojson": "3.0.0-beta.5",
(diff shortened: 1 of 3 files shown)


rreusser (approved):
Yikes. Approve. That shouldn't happen 😞 👎

LyrickCZE:
Question - when can I expect this fix to be live at geojson.io web?
I dont really understand GitHub and all your technical terms tbh : )

Thank you for your help and understanding.

chriswhong:
@LyrickCZE The changes are live now.

LyrickCZE:
You are awesome!
I just tried and can confirm it working same as before.
Thank you very much for your time and very fast fix!

PR #807, opened 2022-12-13
restore console api, update API.md to document it

### LINKED ISSUE #806: API to programmatically access map / data?
Hi,

first of all - I'm very happy that the development of geojson.io has been continued. However, as I understand the underlying architecture has been changed from leaflet to mapbox, I have few question related to rather specific use case of mine.

Previously, I had written a tampermonkey add-on to enable few additional functionality automatically/programmatically on top of geojson.io, which more or less relied on either using api.data.mergeFeatures to append programmatically generated new geojson features both to data object as well as on map or map to add features (e.g. L.circle) or modify manually created features (e.g.  modify key-value pairs for all features).

More specifically, I had few use cases which consisted of  
(1) button asking user entering coordinates which then added a new point feature to specific location (ie generated appropriate geojson point + appended that on map)
(2) button to ask a set of coordinates from user which then generated set of point features to data object + appended those on map
(3) assigning specific key-value pair to all geojson features
(4) programmatically generating certain overlay features (one example being  L.circle with specific radius around all point features, *without* appending these to the data model to simply toggle this overlay on/off)
(5) programmatic generation of all possible points within a hand-drawn region, with certain interval using turf and exporting generated data set (with/without impacting data model)

While I understand these are  highly specific features and not likely needed for general usage, I'm interested to understand whether it is still possible somehow to port my old code against the new implementation, ie. is there a way to access (from Javascript) both map and data directly (other than just thru URL parameters)?

chriswhong:
Restores the geojson.io `console api` with `map`, `draw`, and `data` properties available on `window.api`.  `API.md` is also restored to include documentation for the console api, with updated links (and references to leaflet and leaflet draw replaced with mapboxgl and mapbox-gl-draw)

Closes #806

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,6 +1,9 @@
 ## Geojson.io API
 
-You can interact with geojson.io programmatically via url parameters:
+You can interact with geojson.io programmatically in two ways:
+
+* [URL parameters](#url-api)
+* [Browser console](#console-api)
 
 ## URL API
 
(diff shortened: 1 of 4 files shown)


wboykinm (changes requested):
I'm getting some issues with rollup that I'll hash out separately, but first I'll mention that we need a CHANGELOG entry on this PR.

chriswhong:
Because we aren't doing real releases and CHANGELOG includes headings for dates, I've been manually updating it after merging PRs.  

I am not sure what the right approach here is, as the person writing the PR may not know if and when it will actually get merged.

wboykinm:
@chriswhong Fair point. Onward!

API.md
```diff
@@ -63,3 +66,49 @@ The url is in the form:
...
+    'Map tiles by <a target="_top" rel="noopener" href="http://stamen.com">Stamen Design</a>, under <a target="_top" rel="noopener" href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a target="_top" rel="noopener" href="http://openstreetmap.org">OpenStreetMap</a>, under <a target="_top" rel="noopener" href="http://creativecommons.org/licenses/by-sa/3.0">CC BY SA</a>'
+})
+
+window.api.map.addLayer({
+    'id': 'simple-tiles',
+    'type': 'raster',
+    'source': 'raster-tiles',
+    'minzoom': 0,
+    'maxzoom': 22
+})
```
wboykinm, line 98: This is a fabulous example, though we should be alert to [the rumblings](https://stamen.com/30daymapchallenge-2022/) about Stamen shutting down their tile endpoints.

image

wboykinm (approved):
Perfect. I wish I had a way off the top of my head to use `window.api.on(event, fn)`, but the docs are a great guide to the feature set here.

API.md
chriswhong: Noted, thanks for bringing that up... we can swap it out later I think.

PR #809, opened 2023-01-12
Speed up the flying animations

dabreegster:
Before, opening a file from the default globe view takes about 5s to zoom in. This feels excessive to me:

https://user-images.githubusercontent.com/1664407/212032226-3ae3b2cd-6780-447d-a586-76278d5b3d1b.mp4

I've capped it in this PR to 1s:

https://user-images.githubusercontent.com/1664407/212032290-15acf646-1850-4e7d-8606-e0a2424019a3.mp4

This feels much more convenient to me, but it is an opinionated design question, so happy for other ideas. Also, the change affects other uses of `zoomextent`, like clicking "Zoom to features", loading WKB text, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ Full API documentation can be found in [API.md](API.md).
 
 `rollup` can take several seconds to build before changes appear in the browser.
 
+If you get an error resolving dependencies related to `rollup` on newer versions of node, then try `npm install --force` and be sure to not commit changes to `package-lock.json`.
+
 ## Production Build & Deployment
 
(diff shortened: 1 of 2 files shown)


chriswhong:
Hi, thanks for this contribution.  I don't have strong opinions on this but agree that waiting several seconds just for a flyTo() animation is distracting, especially if you came to edit and want to move quickly.  1s seems pretty abrupt and causes a "blank" map while tiles are loading but at least you get some kind of idea of what part of the world you are flying to if you start from a the zoomed out globe.  

I'll get this merged now, we can always change it to a longer delay if users find it to be too abrupt.

chriswhong (approved):

PR #811, opened 2023-01-16
update external geojson example link

chriswhong:
Updates the example link for the `data=data:text/x-url` url parameter.  The old link was from a legacy mapbox api call that now returns an error message instead of geojson.

The new example uses a geojson file hosted on github for counties in the U.S. state of California.


Countdown369 (approved):

Countdown369:
LGTM!

PR #819, opened 2023-02-16
calculate geometry info based on the full feature

### LINKED ISSUE #816: The Sq. Kilometers value reported from Info popup is incorrect
**Steps to reproduce:**
1. On a macOS, connect to https://geojson.io/
2. On the right pane, paste the following geojson content:
```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              172.2407624558061,
              -42.22493673282945
            ],
            [
              172.2002534098677,
              -42.36145516200265
            ],
            [
              172.47712768621062,
              -42.40644499819358
            ],
            [
              172.51637963704528,
              -42.26940852518235
            ],
            [
              172.2407624558061,
              -42.22493673282945
            ]
          ]
        ]
      }
    }
  ]
}
```
3. Select Meta -> Zoom to features.
4. Mouse click any area within the geometry on the map.
5. Properties/Info popup shows up.  Switch to the Info tab.

**Current Behavior:** The Sq. Kilometer value is returning a random number.  The Sq. Kilometer value could be returned as 300.17 sqkm, 127.05 sqkm, or 95.77 sqkm, depending on which part of the geometry area I clicked on the map.  If I zoom out, the Sq. Kilometer value may return correctly.

I'd expect that the Sq. Kilometer value returns as **~362 sqkm**, with or without Zoom in/Zoom out.

chriswhong:
Fixes #816 by fetching the full feature from the datastore before doing geometry calculations.  (use the feature returned by queryRenderedFeatures() only for getting the feature id, fetch the full feature by its id, then perform calculations.

### DIFF
--- a/src/ui/map/util.js
+++ b/src/ui/map/util.js
@@ -172,7 +172,11 @@ function geojsonToLayer(context, writable) {
 }
 
 function bindPopup(e, context, writable) {
-  const [feature] = e.features;
+  // build the popup using the actual feature from the data store,
+  // not the feature returned from queryRenderedFeatures()
+  const { id } = e.features[0];
+  const feature = context.data.get('map').features[id];
+
   const props = feature.properties;
   let table = '';
   let info = '';


Countdown369 (approved):
LGTM

PR #834, opened 2023-03-29
add ga4 tracking

chriswhong:
Adds a script tag for the new Google Analytics 4 tracking.  Leaves the old GA tracking script in place, which we can remove once the service is retired.


Countdown369 (approved):
Just adds this tag, LGTM.

PR #839, opened 2023-04-13
Fixed a bug when editing geometry in OSM style

### LINKED ISSUE #812: User cannot edit geometries when OSM style is selected
Editing does not work if you have selected the map material from OSM. The path points cannot be selected.

kaitok:
Fixes #812 

When using text-field for symbold in mapbox-gl-js, glyphs is required.
https://github.com/mapbox/mapbox-gl-js/blob/ec307c74bfe46f02e2b6f7292332caa5779a3f2e/src/style-spec/validate/validate_property.js#L54

### DIFF
--- a/src/ui/map/styles.js
+++ b/src/ui/map/styles.js
@@ -24,6 +24,7 @@ module.exports = [
     style: {
       name: 'osm',
       version: 8,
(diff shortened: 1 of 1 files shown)


chriswhong:
Thank you!  Great find!

PR #840, opened 2023-04-14
Fix import bug when missing geometry

### LINKED ISSUE #713: Importing Point FeatureCollection with missing geometry throws alert
Test by saving the geojson below into a file and importing via the Open menu.

This is valid JSON but not valid GeoJSON.  It's getting caught in `addMarkers()` when trying to check the geometry type, and shows an alert "Invalid JSON"

What should happen:
This should be loaded and populated in the JSON editor, but should show an error in the editor as invalid geojson.

```
{
    "type": "FeatureCollection",
    "features": [
        {
            "type": "Feature",
            "properties": {},
            "geometry": {
                "coordinates": [
                    -21.420316844911667,
                    58.589172553022394
                ],
                "type": "Point"
            }
        },
        {
            "type": "Feature",
            "properties": {}
        }
    ]
}
```

kaitok:
Fixes #713 
Features that do not include geometry are not displayed on the map as markers and can be imported.
The geojson will be displayed in the right panel, same as when you copy and paste.

### DIFF
--- a/src/lib/zoomextent.js
+++ b/src/lib/zoomextent.js
@@ -4,7 +4,7 @@ module.exports = function (context) {
   const geojson = context.data.get('map');
   // if the data is a single point, flyTo()
   if (
-    geojson.features.length === 1 &&
+    geojson.features.filter((feature) => feature.geometry).length === 1 &&
     geojson.features[0].geometry.type === 'Point'
   ) {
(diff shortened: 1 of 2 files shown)


chriswhong:
Works great!  Thanks for the fix!

PR #842, opened 2023-04-18
Marker maki symbols

razdravan:
Added the options to add maki symbols to markers.


src/ui/map/util.js
```diff
@@ -113,9 +126,16 @@ const addMarkers = (geojson, context, writable) => {
...
+    if (d.properties && d.properties['marker-symbol']) {
+        symbol = d.properties['marker-symbol']
+    }
+
     const marker = new ClickableMarker({
       color,
-      scale
+      scale,
+      symbol,
+      symbolColor
```
chriswhong, line 138: `symbol-color` is not included in the `simplestyle-spec` https://github.com/mapbox/simplestyle-spec/tree/master/1.1.0

I'm fine adding it, but we should document this.  Can you add a section to `help.md` called "I want to style features" mentioning that geojson.io supports an extended version of simplestyle-spec, and explain what all the options are, and link to the simplestyle-spec repo?

chriswhong (changes requested):
Hi, thanks for this contribution! 

We will need to add a script (or maybe there is rollup config that can do it) to get the maki icon svgs into the build, and add some documentation.

I think I noticed a bug while testing this, the datalist dropdowns in the popups don't seem to be working anymore, but I will deal with that separately.

chriswhong:
Two more things to do:

[ ] validate the `marker-symbol` property, only display a maki symbol if it is a valid maki icon name.  Likewise, only show a letter number if it matches a-z/0-9
[ ] Make sure we can show `0`(I left a note in the discord about how to ensure it makes it through to ClickableMarker
[ ] Once https://github.com/mapbox/geojson.io/pull/848 is in, please make sure `npm run lint` and `npm run format` don't return any errors.

PR #848, opened 2023-05-24
update lint command

chriswhong:
Updates the npm lint command in package.json to check all js files.  Also updates the `format` and `lint-staged` commands


Countdown369 (approved):
LGTM!

PR #852, opened 2023-06-02
fix check for marker-symbol property

### LINKED ISSUE #851: geojson with null properties throws an error
http://geojson.io/#data=data:application/json,%7B%22type%22%3A%22Feature%22%2C%22geometry%22%3A%7B%22type%22%3A%22Point%22%2C%22coordinates%22%3A%5B-73.796904%2C42.708105%5D%7D%2C%22properties%22%3Anull%7D

```
Cannot read properties of null (reading 'marker-symbol')
    at util.js:130:23
```
Looks like we are checking for `marker-symbol` without making sure properties exists first.

chriswhong:
Small fix to check for the existence of properties before checking for `marker-symbol` property in `util.js`
Fixes #851

### DIFF
--- a/src/ui/map/util.js
+++ b/src/ui/map/util.js
@@ -17,8 +17,8 @@ const markers = [];
 makiNames = require('@mapbox/maki/layouts/all.json');
 let makiOptions = '';
 
-for (var i = 0; i < makiNames.length; i++) {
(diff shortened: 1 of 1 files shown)


mappingvermont (approved):
🌮

PR #855, opened 2023-06-20
Fix OSM tile URL

otbutz:
see https://github.com/openstreetmap/operations/issues/737


chriswhong (approved):
Thanks for your contribution!

PR #856, opened 2023-06-27
bump geojsonhint to 3.2.0

### LINKED ISSUE #849: Presence of CRS element should not block display
There are tools around, including GDAL's ogr2ogr (at least as recently as 3.0.4) which produce GeoJSON with this element at the top level

> "crs": { "type": "name", "properties": { "name": "urn:ogc:def:crs:OGC:1.3:CRS84" } },

geojson.io refuses to validate any file with this element, showing this error:

image

That means you can't display the output of a GeoJSON file produced by GDAL. Even though the file is perfectly valid - it's simply a deprecated attribute that should be ignored.

chriswhong:
Updates geojsonhint to v3.2.0 which ignores a top-level crs property.  Closes #849

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,9 +13,10 @@
         "@mapbox/geojson-extent": "^1.0.1",
         "@mapbox/geojson-normalize": "0.0.1",
         "@mapbox/geojson-rewind": "0.5.2",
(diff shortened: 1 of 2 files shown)


Countdown369 (approved):
Basic dependency upgrade, LGTM

PR #875, opened 2024-01-31
bump mapbox gl js to v3, replace Streets with Standard

chriswhong:
image


package-lock.json
OdNairy, line 1: Is it expected that `npm ci` command fails without `--force` flag?

OdNairy: CleanShot 2024-01-31 at 09 48 50@2x

OdNairy (approved):
Thanks @chriswhong for an update!

wboykinm (commented):
It feels like a good sign for the dev experience if this is what's required to go from v2 to v3 and streets to standard. The modularity is really straightforward. 🤘

package-lock.json
chriswhong: I remember the same gl-geocoder dependency issue popping up when v3 came out.  Yes, this is expected until we bump the version on mapbox-gl-geocoder.  I'll get to it soon along with some other dependencies.

PR #877, opened 2024-01-31
handle streets style in storage

chriswhong:
Handles the case where `Streets` is kept in localstorage as the active style, causing a blank map because Streets has been replaced with Mapbox Standard.


pszhuchkov (approved):
LGTM

PR #878, opened 2024-01-31
fix streets storage handling

chriswhong:
Fixes a previous fix to handle `Streets` being stored in localstorage as the active style.


pszhuchkov (approved):
lgtm

PR #881, opened 2024-02-06
Fix WKX related functions

lokkju:
The WKX version current used for WKB/WKT/etc functionality has been broken due to it's dependency on Buffer

This pull request updates the WKX library to the newest version, and adds a Buffer polyfill as required

*note: I'm not sure which version of npm to use to generate package-lock.json; the lockfileVersion has changed, and I may need to regenerate it before merging*


chriswhong (approved):
Thanks for this contribution!

lokkju:
Thanks for getting it merged!

PR #916, opened 2025-05-26
open multiple files at once

tnkuehne:
Hey, thanks for the great tool.

I have the use case that I often need to open multiple files at once. At the moment I need to use the open dialog multiple times to achieve this. With this change I can select multiple files in the dialog which are then opened.

## What's changed
* add multiple attribute to input tag
* iterate over selected files using existing file read behavior

Tested on Ubuntu 24.04 in Firefox and Chromium with up to 10 geojson files and it worked without any problem.

### DIFF
--- a/src/ui/file_bar.js
+++ b/src/ui/file_bar.js
@@ -232,19 +232,22 @@ module.exports = function fileBar(context) {
         .select('body')
         .append('input')
         .attr('type', 'file')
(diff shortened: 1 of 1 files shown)


AndrewSepic (approved):
@tnkuehne Thanks for the submission.   Works great! Happy to merge this.

PR #924, opened 2025-06-17
Replace classic styles with Standard

chriswhong:
This PR replaces the following classic styles:
- `satellite-streets` is replaced with `standard-satellite`
- `light-v11` is replaced with `standard`, with `theme` 'monochrome' 
-  `dark-v11` is replaced with `standard`, with `theme` 'monochrome' and `lightPreset` set to `night`

For all styles that use Mapbox Standard, `show3dObjects` is set to false so that 3d buildings will not interfere with drawn layers.

Logic for changing the color of overlaid geojson feature layers is updated accordingly (Satellite shows green features, dark shows light gray features)

The only remaining classic style is `outdoors-v12`

### DIFF
--- a/src/ui/layer_switch.js
+++ b/src/ui/layer_switch.js
@@ -21,9 +21,11 @@ module.exports = function (context) {
       // this will likely run before the initial map style is loaded
       // streets is default, but on subsequent runs we must change styles
       if (context.map._loaded) {
-        const { title, style } = d3.select(clicked).datum();
+        const { title, style, config } = d3.select(clicked).datum();
 
-        context.map.setStyle(style);
+        context.map.setStyle(style, {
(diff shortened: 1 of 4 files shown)


AndrewSepic:
**Update:**. I believe this is a bug in the logic setting the style somewhere.. 

```
 // handle previous users who had Streets selected
    if (activeStyle === 'Streets') {
      activeStyle = 'Standard';
    }

    const { style, config } = styles.find((d) => d.title === activeStyle);
    ```
    
    In this block of code, `activeStyle` === `Dark`. And that does not match any styles `title` value, so this destructuring is returning undefined.
    
    
    TO test, try leaving the style on the dark theme, and reload the app.

AndrewSepic (approved):
Approved!

PR #943, opened 2026-01-16
fix deckgl circle offset when using standard style

chriswhong:
image

This PR fixes a bug where the vertices rendered on the map are misaligned with other map features, only when using the Mapbox Standard Style or one of its configured derivatives, only at certain zoom levels (observed around z12.5) and only when the vertices are rendered away from the center of the map.

The vertices are drawn using `MapboxOverlay` from `@deck.gl/mapbox`. On investigation, I discovered the deck viewport zoom was not the same as `map.getZoom()` and that this mismatch was caused by map `terrain` which is enabled by default in the Mapbox Standard Style. The fix is to disable terrain as soon as a style loads using `map.on('style.load')`


AndrewSepic (approved):
🚢   ✨

PR #958, opened 2026-02-27
Fix extra coordinate when creating polygons

AndrewSepic:
This PR fixes a bug when creating polygons and terminating the drawing with a double click.  Previously when closing the polygon with a doubleclick there was an additional duplicate coordinate created.  ie: a 4 sided polygon had 6 coordinates instead of 5.  

This fixes the extra coordinate, properly handling how browsers register clicks.  When double clicking on the Map 3 points are actually created, therefore the `splice()` method needed to remove an additional point.  The `line.ts` handler (which works as expected) already follows this pattern.


chriswhong (approved):
LGTM, works as expected.

PR #961, opened 2026-04-01
cw/simplestyle markers

chriswhong:
Adds support for `simplestyle-spec` properties for point features in geojson.io/next. Covers both rendering of simplestyle spec point properties and adds controls for choosing them in the **Styles** menu in the **Feature Editor**

image

## `marker-size` 

`small` renders a marker at 4px radius, the same as default markers in geojson.io/next.
`medium` and `large` have radius 8px and 12px respectively.

## Styles Panel

The styles panel is update to only show relevant simplestyle-spec inputs for the selected feature's geometry type, (point features do not need `fill-` or `stroke-` properties, so they are not shown, etc). 

`marker-color` shows the color picker
`marker-size` shows a dropdown with the three valid values
`marker-symbol` shows a custom popover where users can click a maki icon or a-z/0-9 character, or search by id. 

image

### DIFF
--- a/next/app/components/panels/feature_editor/feature_editor_style.tsx
+++ b/next/app/components/panels/feature_editor/feature_editor_style.tsx
@@ -2,24 +2,33 @@ import { ColorPopoverField } from 'app/components/color_popover';
 import {
   Button,
   inputClass,
+  PopoverContent2,
   StyledLabelSpan,
   styledCheckbox,
   TextWell
 } from 'app/components/elements';
 import { useAutoSubmit } from 'app/hooks/use_auto_submit';
 import { purple900 } from 'app/lib/constants';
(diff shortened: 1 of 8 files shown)


AndrewSepic (changes requested):
This is working well and I think it's a great feature.  Currently when you create a new point, it is automatically added with a `marker-size` property. 

```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {
        "marker-size": "medium"
      },
      "geometry": {
        "type": "Point",
        "coordinates": [
          -73.1531973,
          44.5110074
        ]
      }
    }
  ]
}
```

We shouldn't be adding properties to users GeoJSON unless they are specifically setting those properties.

AndrewSepic (approved):
Thanks!  LGTM!

PR #965, opened 2026-04-29
Adds collapsible UI to Feature Editor

AndrewSepic:
## Summary 
This PR introduces a 'collapsed' or 'minimized' state to the Feature Editor Panel.  Based on some user feedback the panel takes up valuable real estate and is not part of their work flows.  This allows a user to set the panel to a 'minimized' state to save map real estate and not interfere with map and geometry interations.  This state works for small screen sizes as well. 

https://github.com/user-attachments/assets/b19e2777-4bd9-4eda-ba35-18db0a540ae9

## Implementation Details
- A new jotai `AtomWithStorage` is created to store the UI state for the user in session storage (this could potentially be set to localStorage to persist this across sessions). 
- A `SizeIcon` from Radix UI is added to the top bar of the Feature Editor (along with descriptive tooltip) to allow the user to trigger collapsed and expanded states
- State is used to derive panel sizing and positioning and CSS transitions enable a smooth visual between the different states.


chriswhong:
Looks great! 

1) What do you think about keeping the minimized feature editor horizontally centered (this is more consistent with how the feature editor is displayed)

2) We are using the `size` radix icon for both collapse and expand actions. The only other place it is used in the app is to open an expanded UI for cells in data tables. I think its use on the collapsed feature editor is consistent, but we should find another option for the collapse option full size feature editor.  We could use chevron down or arrow down, or make a custom icon that resembles `size` but with the arrows pointing inward.

chriswhong (approved):
Looks great! I think a keyboard shortcut may be in order to expand and collapse this, but we can handle that later.

PR #966, opened 2026-05-05
Support legacy hash-based #data= URL params

chriswhong:
## Summary                                                                
                                                            
  Adds support for the legacy hash-based URL formats used by older versions 
  of geojson.io, and cleans up the URL after data loads.
                                                                            
  **Legacy hash formats supported:**                        
  - `#data=data:application/json,<encoded-geojson>` — inline GeoJSON
  - `#data=data:text/x-url,<encoded-url>` — remote file via URL             
                                                                            
  In both cases the data is loaded and the hash is cleared from the URL on  
  success. The existing `?data=` and `?id=` query params also have their    
  params stripped after a successful load.                                  
                                                            
  **Before:** Legacy URLs would fail silently or not load at all. Query     
  param URLs would persist in the address bar after loading.
                                                                            
  **After:** All URL formats load correctly, and the address bar is clean   
  after data loads. Failed loads leave the URL intact so the user can see
  what went wrong.                                                          
                                                            
  ## Test plan

  - [x] `#data=data:text/x-url,https%3A%2F%2Fraw.githubusercontent.com%2Fcod
  eforgermany%2Fclick_that_hood%2Fmain%2Fpublic%2Fdata%2Fcalifornia-counties
  .geojson` — loads remote GeoJSON, hash disappears                         
  - [x] `#data=data:application/json,%7B%22type%22%3A%22LineString%22%2C%22c
  oordinates%22%3A%5B%5B0%2C0%5D%2C%5B10%2C10%5D%5D%7D` — loads inline      
  GeoJSON, hash disappears
  - [ ] `?data=<url>` — loads data, `?data=` param is removed after load    
  - [x] `?id=gist:...` — loads gist, `?id=` param is removed after load     
  - [x] Failed loads leave the URL unchanged

### DIFF
--- a/next/app/components/url_api.tsx
+++ b/next/app/components/url_api.tsx
@@ -52,7 +52,6 @@ export function UrlAPI() {
     if (!map) return; // Wait for map to be available
 
     // Handle legacy hash-based data params (e.g. #data=data:application/json,... or #data=data:text/x-url,...)
-    // Normalize to canonical ?data= query param and update the URL
     let legacyData: string | null = null;
     const hashString = window.location.hash;
     if (!data && !id && !gist && hashString.startsWith('#data=')) {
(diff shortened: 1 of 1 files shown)


AndrewSepic (approved):
Works nicely!   Thanks. 🚢

PR #967, opened 2026-05-05
Add `map` query param to persist camera view

chriswhong:
## Summary
                                                                       
  Implements a ?map=zoom/lat/lng[/bearing[/pitch]] query parameter that
  keeps the URL in sync with the map camera at all times and restores the
   camera position on load.
                                                                         
  Format                    

  ?map=zoom/lat/lng
  ?map=zoom/lat/lng/bearing
  ?map=zoom/lat/lng/bearing/pitch

  - bearing and pitch are optional (default to 0 when omitted)           
  - Values are rounded to reduce URL noise (zoom: 2 decimals, lat/lng: 2
  decimals, bearing: 1 decimal, pitch: integer)                          
  - When combined with ?data or ?id, the map param takes priority over
  flying to the data extent                                              
                            
  Sample URLs to test                                                    
                            
  Basic zoom/lat/lng (NYC):                                              
  https://geojson.io/#map=12/40.72/-73.99
                                                                         
  With bearing (rotated 45°):                                            
  https://geojson.io/#map=12/40.72/-73.99/45
                                                                         
  With bearing + pitch (tilted view of San Francisco):                   
  https://geojson.io/#map=14/37.77/-122.43/30/45                         
                                                                         
  Combined with ?id (map param overrides fly-to):                        
  https://geojson.io/#map=5/39/-98&id=github:...
                                                                         
  Changes                                                                
  
  - parse_map_param.ts — parsing and serialization logic with validation 
  - parse_map_param.test.ts — unit tests covering edge cases and omission
   rules                                                                 
  - map_component.tsx — syncs camera changes to the URL and reads param
  on init                                                                
  - url_api.tsx — passes map param through to map initialization
  - pmap/index.ts — exposes camera event for URL sync                    
  - about.md — documents the new query parameter


next/app/components/map_component.tsx
```diff
@@ -166,8 +197,8 @@ export const MapComponent = memo(function MapComponent({
     if (!map?.map || hasZoomedRef.current) return;
 
     const handleLoad = () => {
-      // Only zoom if there are features (loaded from sessionStorage)
-      if (featureMap.size > 0) {
+      // Only zoom if there are features and no saved camera to restore
+      if (featureMap.size > 0 && !initialCameraRef.current) {
```
AndrewSepic: This check does not work.  ie:  The camera continues to zoom when features are added via url params regardless of whether there is a saved camera state.

AndrewSepic (commented):
I'm wondering if we shouldn't hold these in query params?  ie: I think users like to share URLs which give a certain view.  This doesn't allow that. I think that specific feature (being able to share a URL with camera position) was requested. 

In addition one of the desired test conditions fails, ie: it will always zoom to features loaded via `data` or `id` params, regardless of a saved view.

next/public/about.md
```diff
@@ -120,10 +120,11 @@ When export GeoJSON, geojson.io always exports a FeatureCollection.
 
 You can load data into geojson.io using URL parameters.
```
AndrewSepic: This pretext here should be updated here..  Maybe..
```suggestion
geojson.io supports URL parameters for adding data to the map and storing the map's location and view settings.
```

AndrewSepic (dismissed):
Looks good!  Thank you!  Just one suggestion for update in the `about.md` file.


**Note:** I was able to inconsistently cause an interesting bug. When I removed the `?map=` query param and reloaded the app.  If there were features on the map, the app might not render them on the map (they still existed in the sidebar) and an additional reload would bring them back onto the map.  I had this happen maybe 2% of the time I did this, total about 6 times in testing lots reloads.  While I don't know why someone would take this action within the app, just want to flag it.

AndrewSepic (approved):
LGTM!

PR #975, opened 2026-05-20
Add globe projection support and 3D features toggle

chriswhong:
## Summary

- Adds Globe and Mercator projection selector to the layers/basemap popover, persisted in sessionStorage
- Adds Show 3D Features toggle (buildings, trees, landmarks, facades) for Mapbox Standard-based styles, grayed out for Outdoors/OSM
- Fixes vertex handle overlay alignment in globe mode: uses a round-trip through `map.project()` → `viewport.unproject()` to correct the per-frame position offset between Mapbox's globe math and deck.gl's GlobeViewport
- Fixes initial style loading race condition where globe projection caused Mapbox to internally load Standard style before the explicit `setStyle()` call; projection is now applied in `onMapStyleLoad` after each style load
- Adds async `setStyle` sequence counter to prevent stale fetches from overwriting newer style loads
- Skips globe vertex correction in Mercator mode
- Default initial camera: zoom 2, center [0, 20]

geojson_io___Powered_by_Mapbox_🔊_and_CHANGELOG_md_—_geojson_io_and_index_html_—_tmp-esm-mapbox


next/app/lib/pmap/index.ts
```diff
@@ -264,6 +277,10 @@ export default class PMap {
     const map = event.target as mapboxgl.Map;
     // disable terrain. If enabled in the style (as in Mapbox Standard style), it causes an alignment bug in the deck.gl overlay
     map.setTerrain(null);
+    // Apply the desired projection after each style load. Setting projection before
```
AndrewSepic: Not sure we need this comment here.  Glad the race condition was removed, but don't think we need to call this out.

next/app/lib/pmap/index.ts
```diff
@@ -480,6 +581,18 @@ export default class PMap {
...
+      for (const property of [
+        'show3dOptions',
+        'show3dBuildings',
+        'show3dTrees',
+        'show3dLandmarks',
+        'show3dFacades'
+      ]) {
+        this.map.setConfigProperty('basemap', property, show3d);
+      }
+      // Projection can change without a full style reload
```
AndrewSepic: Should Probably remove this comment.

AndrewSepic (changes requested):
Overall this is working pretty well.  Glad we could find a solution to the problem. Also, happy to learn that Deck.GL has an experimental `GlobeView` which helps bridge the gap.  I made a few comments/suggestions to remove comments I feel are not needed. There may be more self explanatory comments which could be removed.

Two issues I want to flag: 

### Dancing points in Globe view
When you are in glob view and are moving the map around we see this drift and wiggle on points as this `_buildScatterplotLayer` is processing the math for the points.  I'm not sure if this is a blocker, but it does feel a like a not complete solution. Or something that isn't ideal.   (See video in slack)



## Can't change projection with classic styles
If the app is currently running a classic style (OSM or Outdoors) you cannot change the projection of the map.  The UI state will update, but the map does not update with the new projection.

bijanps:
perfect

next/app/lib/pmap/index.ts
chriswhong: updated to a more concise comment:  `// set projection to last known value (if any) so that style reloads don't reset it to default`

chriswhong:
Changes in ddae60591ca6bf260678169cd0d4054c8bd5ba39

**Fix projection changes for OSM/Outdoors styles**                                                                   
                                                                                                                    
  When using OSM or Outdoors as the basemap, toggling the projection (globe ↔ mercator) had no effect. The root     
  cause: setStyle() uses Mapbox's default diff: true, which compares the old and new style JSON. Since projection is
   never embedded in the style JSON (it's applied separately via setProjection), Mapbox saw no diff, skipped the
  reload, and style.load never fired — so setProjection in the onMapStyleLoad handler never ran.                    
                                                                      
  Fix: for non-import styles with only styleOptions changed, call setProjection directly (mirroring what already    
  happened for import-based styles). If only the projection changed, return early with no style reload needed. If
  labels or 3D settings also changed, fall through to setStyle({ diff: false }) to guarantee a full reload.         
                                                                      
  **Fix "Show 3D features" toggle enabled incorrectly for Standard Satellite **                                         
  
  I noticed that the "Show 3d features" checkbox was enabled when Standard Satellite was selected, which should not be the case.
  
  The toggle used json.imports as a proxy for 3D support, but Standard Satellite also uses imports — it just doesn't
   expose 3D feature config (no buildings/trees/landmarks). Added an explicit supports3dFeatures boolean to
  StyleConfigTemplate and set it only on Standard, Standard Light, and Standard Dark.       
  
  **No fix for the vertex jitter during globe view**
  
  I don't think this is a blocker and we can revisit it later if we want to prioritize it.

AndrewSepic (approved):
Thanks for wrapping this up.  Looks good! 👍

PR #992, opened 2026-06-04
GL JS Updates - Revert Light & Dark theme to Classic Styles (Performance)

AndrewSepic:
## Summary
This PR updates to the latest version of GL JS, Reverts 'Standard Light' and 'Standard Dark' import based styles to classic Light/Dark. And address some state update logic which was causing a full map reload on classic styles when toggle labels on/off.

## GL JS Update
Updates to the latest GL JS Version `3.24.0`. which includes a bunch of performance upgrades.

## Removal of Standard light and standard dark styles for Classic Styles
After internal discussion with GL JS team around performance benefits of the 'classic' styles Standard Light and Standard Dark are replaced with Light (`mapbox://styles/mapbox/light-v11`) and Dark (`mapbox://styles/mapbox/dark-v11`) with Light being the default style when the app is loaded on a cold cache.

## Small fix on style logic
Toggling map labels on/off on 'classic' styles was causing a full map reload, when that is not necessary. Some updates to that logic ensure that the map need not fully reload to toggle visibility of labels.


package.json
```diff
@@ -139,11 +139,10 @@
     "zod": "^3.25.76"
   },
   "devDependencies": {
-    "@mapbox/mapbox-gl-style-spec": "^14.17.0",
+    "@mapbox/mapbox-gl-style-spec": "^14.24.0",
     "@types/d3-dsv": "^3.0.7",
     "@types/deep-freeze": "^0.1.5",
     "@types/lodash": "^4.17.20",
-    "@types/mapbox-gl": "^3.4.1",
```
AndrewSepic, line 146: `types/mapbox-gl` is a deprecated package as `mapbox-gl` ships its own types now.

chriswhong (approved):
Looks great, nice catch on the unnecessary style reload when toggling labels.

PR #993, opened 2026-06-04
Feature updates based on User Feedback

### LINKED ISSUE #978: New geojson.io is missing "Clear" feature
Old geojson.io has a menu item to clear the map. If the new geojson.io has this, I haven't found it.

Image

See #371

### LINKED ISSUE #982: New geojson.io numpad "Delete" doesn't work
* Select any feature
* (With numpad disabled), press the `Del` keyboard key
  * Nothing happens
* Press the `Delete` keyboard key
  * Feature is removed

My guess would be it's checking something like `event.code === 'Delete'` rather than `event.key === 'Delete'`

### LINKED ISSUE #988: Missing foldGutter in new Version
## Missing foldgutter in code editor

The previous version of geojson.io included a fold gutter in the code block, which was especially useful for collapsing geometry sections in large GeoJSON files.

This feature was not carried over to the new app (geojson_editor.tsx). Based on a quick look at the codebase, it seems like it just needs to be re-added in the extensions useMemo. According to codemirror docs [`foldGutter()`](https://codemirror.net/docs/ref/#language.foldGutter) extension is already part of `@codemirror/language`.

**Old behavior:** Fold arrows visible in the gutter, allowing geometry blocks to be collapsed.

**Current behavior:** No fold gutter present.

**Suggested fix:** Add `foldGutter()` (and [`foldKeymap`](https://codemirror.net/docs/ref/#language.foldKeymap)) to the CodeMirror extensions.

Screenshot showing old fold gutter

AndrewSepic:
### Delete all / Clear  features button
adds a trash icon button to the feature editor header that  deletes all features on the map in a single click. Uses the same atomic state update and undo moment as the keyboard delete flow, so the operation is fully undoable. A toast notification  reminds the user they can undo with Ctrl-Z.

Screenshot 2026-06-04 at 2 34 19 PM

Closes #978 

### Numpad Delete key fix 
the Delete key on the numpad (numlock off) was not triggering feature deletion. Root cause: react-hotkeys-hook defaults to matching on event.code (physical key position), and the numpad Delete has code NumpadDecimal rather than Delete. Fixed by passing  useKey: true for the delete binding so it matches on event.key (logical value) instead — both keys produce key: "Delete" when numlock is off.

Closes #982 

### Increased map URL coordinate precision
Lat/lng in the ?map= query param were rounded to 2 decimal places (~1km precision). Increased to 4 decimal places (~11m) to match the precision of  the legacy app's Mapbox GL hash. Reported by users who regularly use Geojson.io to debugging geometry in multiple tools which rely on this coordinate precision
 
### CodeMirror fold gutter — 
Adds a `foldGutter()` extension to CodeMirror affecting the GeoJSON editor so users can collapse nested  objects and arrays in the JSON panel.

Closes #988

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 Recent changes to the geojson.io application. Entries are grouped by ship date.
 
+## 2026-06-04
+
+- **Delete all features:** Adds a delete all button to the feature editor header. Operation is undoable via Ctrl-Z. ([#993](https://github.com/mapbox/geojson.io/pull/993))
+- **Numpad Delete key fix:** The Delete key on numpads (numlock off) now correctly deletes selected features. ([#993](https://github.com/mapbox/geojson.io/pull/993))
+- **Map URL coordinate precision:** Increased lat/lng precision in the `?map=` query param from 2 to 4 decimal places. ([#993](https://github.com/mapbox/geojson.io/pull/993))
+- **GeoJSON editor fold gutter:** Adds a fold gutter to the GeoJSON editor panel for collapsing nested objects and arrays. ([#993](https://github.com/mapbox/geojson.io/pull/993))
+- **GL JS Update:** Bump to `v3.24.0` [#992](https://github.com/mapbox/geojson.io/pull/992)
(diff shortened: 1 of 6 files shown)


chriswhong (approved):
All working as expected, aside from the numpad Delete feature which I can't test. The approach looks right so let's see how it goes.

PR #996, opened 2026-06-08
Fix Firefox  Issues

### LINKED ISSUE #987: Can not delete point (vertex) of a polygon
"Editing Geometries

Select a feature to edit its geometry.

...
    Delete to delete a selected vertex.
..."

On polygon vertex can not to be selected. Delete button delete whole feature.

AndrewSepic:
## Summary
The changes here address issues that Firefox browser exposed but which were latent in the application.

### Vertex selection on polygons
Sort comparator missing equal case caused non-deterministic ordering in Firefox causing the inability to select Vertices -  vertices now reliably sort before features when `at` values (feature ids) are equal.  Closes #987 

### 🚨Map blank on fresh load (default style) 
The `onlyStyleOptionsChanged` optimization was firing before the style was applied to the map, causing the in-
flight style fetch to be discarded and leaving the canvas blank; guarded by checking `features` source exists first. (To replicate, load app in Firefox, select 'LIGHT' map style and reload)

### `queryRenderedFeatures` on unloaded style 
Backwards `isStyleLoaded()` condition in `fastMovePointer` was querying layers before the style was ready causing the folllowing errors in the console. Check added to ensure layers source exists before calling `queryRenderedFeatures`.
Screenshot 2026-06-09 at 9 20 20 AM

### Maki icon already exists errors
This exists because maki icons were being added to the sprite sheet to light and dark themes (which already contain maki). Maki icons added to the style (for use as Simple Style Spec rendering) have been added with the `maki-` namespace to remove collision with existing icons and allow for coloring in map rendered simple style.
Screenshot 2026-06-09 at 9 31 04 AM

All four bugs were reproducible in Firefox; the `maki` icons issue and `queryRenderedFeatures` issues also produced silent errors in Chrome.

## Test plan

- [x] Fresh reload with LIGHT style in Firefox — map renders
- [x] Fresh reload with DARK style in Firefox — map renders  
- [x] Click a polygon vertex in Firefox — vertex selected, Delete removes vertex not whole feature
- [x] Toggle styles in Firefox — all styles load correctly
- [x] Hover over features — cursor changes correctly
- [x] Use simple style spec with marker-symbol - icons render after style load

### DIFF
--- a/app/components/map_component.tsx
+++ b/app/components/map_component.tsx
@@ -4,7 +4,10 @@ import { env } from 'app/lib/env_client';
 import type { FlatbushLike } from 'app/lib/generate_flatbush_instance';
 import { EmptyIndex } from 'app/lib/generate_flatbush_instance';
 import { useHandlers } from 'app/lib/handlers/index';
-import { CLICKABLE_LAYERS } from 'app/lib/load_and_augment_style';
+import {
+  CLICKABLE_LAYERS,
+  FEATURES_SOURCE_NAME
+} from 'app/lib/load_and_augment_style';
 import { wrappedFeaturesFromMapFeatures } from 'app/lib/map_component_utils';
 import type { PMapHandlers } from 'app/lib/pmap';
 import PMap from 'app/lib/pmap';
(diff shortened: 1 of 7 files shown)


chriswhong (approved):
🚢

PR #1006, opened 2026-06-11
Streamline import: skip modal for non-ambiguous file types

chriswhong:
## Summary

This PR streamlines imports, removing the need for the modal for most import types. For most imports, the modal was only necessary for toggling the "Remove coincident vertices" setting. The ability for the user to set this has been removed from the UI for now, but the state handlers for it remain in place if we ever want to re-introduce it.

- **Skip the import modal** for file types that don't need user input (geojson, kml, topojson, shapefiles, etc.). The modal now only appears for `csv`, `xls`, and `coordinate-string` where column mapping or coordinate order is ambiguous.
- **Extract `useImportFiles` hook** so drag-and-drop (`drop.tsx`) and the file picker (`use_open_files.ts`) share identical classification, auto-import, and `fitBounds` logic.
- **Fix `fitBounds` on drag-and-drop** — map now flies to imported data extent after dropping a file.

## Test plan

- [ ] Drop a GeoJSON file — modal should NOT appear, data imports and map flies to extent
- [ ] Drop a shapefile — same, no modal
- [ ] Drop a CSV — import modal should appear with column options
- [ ] Use Import button → same behavior as drag-and-drop for each file type
- [ ] Toggle "Remove coincident vertices" in Options → setting persists after page reload
- [ ] Verify the setting is applied during import (both drag-and-drop and modal paths)


AndrewSepic:
I'm not sold on the idea of adding a new **Options** UI menu item to handle the  "Remove coincident vertices"  toggle.  

I am not entirely clear on the different import types here and what they look like. Can you provide some testing files that demonstrate how these are handled currently? 

If we can import geojson files without a modal, that's great and I think that should be the core fix here, but adding more UI to handle app level constant without a way to test and add compare current branch to this branch is a stretch to include.

chriswhong:
[geojson.io sample data.zip](https://github.com/user-attachments/files/28961100/geojson.io.sample.data.zip)
This is a local collection of test files, including CSV, OSM, SHP, KML, KMZ, GPX, encoded polyline, etc.

chriswhong:
My reasoning here is there are bound to be more app-level config settings in the future as this project grows, so a UI for that is going to be necessary one day.

If we want to allow users to specify whether the app should remove coincident vertices, that UI must live somewhere. I think it makes sense to remove them as a default, but I can see some users not wanting geojson.io to alter their data (I'm sure that's why this was made optional to begin with)

chriswhong:
Updated to remove the global options modal. The remove coincident vertices option remains in state set to false in case we ever want to enable it in the future, but for now all imports will not touch coincident vertices.

AndrewSepic (dismissed):
LGTM!  Thanks for refining this PR.

AndrewSepic (approved):
🚢

PR #1007, opened 2026-06-11
Add vertex-snap mode: Shift+Option+drag snaps to vertices with highlights

chriswhong:
## Summary

Improves snapping by providing a modifier key for snapping to vertices only, not edges. Highlights vertices on nearby features. 

More on the use case: I often need to hand-draw a polygon adjacent to an existing polygon, and want them to share vertices. Snapping is very helpful here, but depending on zoom the current snapping feature sometimes gets you close but not quite on the adjacent feature's vertex. There are also situations where many vertices make a smooth curve and it may not be clear where the vertices are, so showing them in a subtle way is helpful for knowing what needs to be added to a new polygon.

vertex snap

- **Option+drag** a vertex snaps to the nearest edge of a nearby feature (existing behavior, unchanged)
- **Shift+Option+drag** a vertex now snaps to the nearest *actual vertex* of nearby features (not the closest point on an edge)
- When in vertex-snap mode, all vertices of features within ~80px of the cursor are **highlighted as amber circles**, giving visual guidance for precisely aligning polygon edges to match existing geometries

## Implementation

- New `EphemeralEditingStateVertexSnap` type in `state/jotai.ts` carries vertex positions to the render layer
- `getSnappingCoordinates` in `utils.ts` accepts a `verticesOnly` flag — when true, snaps to the nearest coordinate vertex using turf `distance()` rather than `nearestPointOnLine()`  
- `getNearbyVertices` in `utils.ts` queries a wider 80px search box and collects all vertices from every nearby feature
- New deck.gl `ScatterplotLayer` in `pmap/index.ts` renders the amber highlight circles when ephemeral state is `'vertex-snap'`
- Ephemeral state is cleared automatically on mouse-up (existing `up` handler)

## Test plan

- [ ] Select a feature with vertices (polygon or line)
- [ ] Hold Option and drag a vertex — should snap to nearest point on nearby edges (existing behavior)
- [ ] Hold Shift+Option and drag a vertex — nearby feature vertices should appear as amber circles; dragging should snap to the nearest highlighted vertex
- [ ] Release mouse — highlights disappear
- [ ] Verify no highlights appear when dragging without modifier keys

### DIFF
--- a/app/lib/handlers/none.ts
+++ b/app/lib/handlers/none.ts
@@ -71,8 +71,6 @@ export function useNoneHandlers({
 
   // Immediately show/hide vertex-snap highlights when modifier keys change,
   // without waiting for the next mousemove event.
-  // Immediately show/hide vertex-snap highlights when modifier keys change,
-  // without waiting for the next mousemove event.
   useEffect(() => {
     const handleKey = (e: KeyboardEvent) => {
       const draggingVertex =
--- a/app/lib/handlers/none.ts
+++ b/app/lib/handlers/none.ts
@@ -14,7 +14,7 @@ import { useEndSnapshot, useStartSnapshot } from 'app/lib/persistence/shared';
 import { captureException } from 'integrations/errors';
 import { useSetAtom } from 'jotai';
 import noop from 'lodash/noop';
-import { useRef } from 'react';
+import { useEffect, useRef } from 'react';
 import { USelection } from 'state';
 import {
   cursorStyleAtom,
@@ -25,7 +25,11 @@ import {
 } from 'state/jotai';
 import { modeAtom } from 'state/mode';
 import type { HandlerContext, IWrappedFeature } from 'types';
-import { getMapCoord, getSnappingCoordinates } from './utils';
+import {
+  getMapCoord,
+  getNearbyVertices,
+  getSnappingCoordinates
+} from './utils';
 
 export const getAllFeatures = ({
   featureMap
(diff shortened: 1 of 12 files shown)


AndrewSepic (commented):
This works with existing polygons, but does not work when creating a new polygon.  ie: creating a new Polygon adjacent to an existing one, does not have the ability to vertex snap, (although snapping to edges does work).  I think this should behaviour should be the same whether you are working on an existing feature or creating a new feature.

AndrewSepic (changes requested):

chriswhong:
Snapping is now working when drawing new linestrings and polygons, and I also made sure that it also works when when drawing the initial vertex of a linestring or polygon, or when drawing a point.

AndrewSepic (dismissed):
Thanks for the fix!  Looking good.  Left a small comment to make sure the change was intentional.

app/lib/handlers/none.ts
```diff
@@ -59,9 +63,48 @@ export function useNoneHandlers({
...
   const lastPoint = useRef<mapboxgl.LngLat | null>(null);
+  const lastMoveEvent = useRef<
+    mapboxgl.MapMouseEvent | mapboxgl.MapTouchEvent | null
+  >(null);
   const spaceHeld = useSpaceHeld();
   const altHeld = useAltHeld();
 
+  // Immediately show/hide vertex-snap highlights when modifier keys change,
+  // without waiting for the next mousemove event.
+  useEffect(() => {
```
AndrewSepic, line 74: With the most recent commits, the vertexes are shown on a geometry when you press `option`, not when you press `option+shift`.  I don't know if that's a bad thing, but it's worth sharing. Maybe a bit strange that you are showing vertexes when you aren't snapping to them..

chriswhong: Confirmed the vertices only show on nearby features when both option and shift are held.

AndrewSepic (approved):
Works!   Thanks for the iterations & attention to detail.

PR #1008, opened 2026-06-11
Add Share button with URL-encoded data link

chriswhong:
## Summary

- Adds a **Share** menu item after Export in the toolbar
- Opens a modal with a shareable URL that encodes the current GeoJSON as a `?data=data:application/json,...` query parameter — no upload or account needed
- Share button is disabled (grayed out) when the resulting URL would be ≥ 8000 characters, with a tooltip explaining why
- A warning is shown if the URL is greater than 2000 characters
- Adds a **Sharing data** section to the About panel documenting the feature and its URL length limit
- Confirmed through testing that geojson.io will return `Error: URI Too Long` and not load the app if the URL is 8220 characters or longer.

## Test plan

- [ ] Draw one or a few small features → Share button becomes enabled
- [ ] Click Share → modal opens with a URL in a read-only input
- [ ] Click copy icon → URL is copied to clipboard, toast shows
- [ ] Open the copied URL in a new tab → same features load on the map
- [ ] Add many features with complex geometry until the encoded URL exceeds 2000 chars → Share button becomes grayed out with tooltip "Dataset too large to share via URL"
- [ ] No features on map → Share button is disabled (same as Export)
- [ ] Check About panel → "Sharing data" section is present

### DIFF
--- a/app/components/dialogs.tsx
+++ b/app/components/dialogs.tsx
@@ -10,6 +10,7 @@ import { RasterLayerDialog } from 'app/components/dialogs/raster_layer';
 import { QuickswitcherDialog } from 'app/components/search/quickswitcher';
 import SimplifyDialog from 'app/components/dialogs/simplify';
 import AboutModal from 'app/components/about_modal';
+import { ShareDialog } from 'app/components/dialogs/share';
 import { VisuallyHidden } from '@radix-ui/react-visually-hidden';
 import { useAtom } from 'jotai';
 import { Dialog as D } from 'radix-ui';
@@ -68,6 +69,7 @@ export const Dialogs = memo(function Dialogs() {
       extraWide = true;
       return <AboutModal open={true} />;
     })
+    .with({ type: 'share' }, () => <ShareDialog onClose={onClose} />)
     .exhaustive();
 
   return (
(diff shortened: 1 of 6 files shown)


AndrewSepic (changes requested):
Can you clarify what the differences are between the 8000 char check and the 2000 char check?  Why are we warning at 2000 chars?  And I assumed with the 8000 char limit browsers just won't parse the URL, ie the app won't load.

I think we should use an icon here to support the sharing activity and reduce the real estate needed for the new item. 

We could use the share icon from the Dialog, or implement a modified icon of the ['paper plane' variety](https://thenounproject.com/browse/icons/term/paper-plane-send/) which is used on social applications and could fit here.

chriswhong:
The 8000 character check represents the actual limit for geojson.io since we know it won't work above 8220 characters. A quick search on the topic shows 2048 as a de facto standard for URL length, though based on older browser restrictions that don't actually apply anymore.

Happy to add an icon instead of a text menu item.

AndrewSepic:
I was testing the limits of the geojson in the app which could be shared and was a litle dissappointed in how little is managed to be fit into the URL encoded geojson.  Claude picked up that there was actually double encoding happening, causing duplicate characters being created. That was removed in a commit I pushed above ☝️ 

And next I created a small PR #194 that uses gzip and base64 encoding to give the app a huge improvement in the number of features on the map which could be encoded into the 8000 character limit. Roughly 90% improvement.  

Would love for you to test that out.  The normal URL encoded data continues to work as a fallback. No breaking changes.

chriswhong:
Superseded by #1014

PR #1010, opened 2026-06-12
Fix draw tool menu layering over mode hints

### LINKED ISSUE #1000: Z-Order on overlapping UI
When interacting with the draw tools there is some adjustment needed in the z-order of tool prompts and drop downs from the tool menu itself. 
This screenshot demonstrates the poor UX. 

Image

zzzhangy:
## Summary
- Keep dropdown and context menu panels on the floating content z-index layer.
- Move map mode hints to a lower layer so draw tool menus render above them.

Closes #1000

## Validation
- `npx prettier --check "app/components/elements.tsx" "app/components/mode_hints.tsx"`
- `npx eslint "app/components/elements.tsx" "app/components/mode_hints.tsx"`
- `git diff --check`
- `npm run build`
- `npm test`
- Browser check: Rectangle mode + Circle type dropdown; dropdown computed `z-index: 50`, mode hint computed `z-index: 10`, and the overlap point resolves to the `Mercator` menu item.

### DIFF
--- a/app/components/elements.tsx
+++ b/app/components/elements.tsx
@@ -338,11 +338,13 @@ export const contentLike = `py-1
     rounded-sm
     shadow-[0_2px_10px_2px_rgba(0,0,0,0.1)]
     ring-1 ring-gray-200 dark:ring-gray-700
-    content-layout z-50`;
+    content-layout`;
 
-export const DDContent = classed(DD.Content)(contentLike);
-export const CMContent = classed(CM.Content)(contentLike);
-export const CMSubContent = classed(CM.SubContent)(contentLike);
+const floatingContentLike = `${contentLike} z-50`;
+
+export const DDContent = classed(DD.Content)(floatingContentLike);
+export const CMContent = classed(CM.Content)(floatingContentLike);
+export const CMSubContent = classed(CM.SubContent)(floatingContentLike);
 
 const styledLabel =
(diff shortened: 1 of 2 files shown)


app/components/elements.tsx
```diff
@@ -338,11 +338,13 @@ export const contentLike = `py-1
     rounded-sm
     shadow-[0_2px_10px_2px_rgba(0,0,0,0.1)]
     ring-1 ring-gray-200 dark:ring-gray-700
-    content-layout z-50`;
+    content-layout`;
 
-export const DDContent = classed(DD.Content)(contentLike);
-export const CMContent = classed(CM.Content)(contentLike);
-export const CMSubContent = classed(CM.SubContent)(contentLike);
+const floatingContentLike = `${contentLike} z-50`;
```
chriswhong, line 343: Can you add a comment here explaining why these three elements need a different z-index?

app/components/mode_hints.tsx
```diff
@@ -23,8 +23,9 @@ function ModeHint({
   return (
     <div
       className={clsx(
-        'z-0 absolute top-2 left-2 px-2 text-sm flex gap-x-2 items-center dark:text-white',
-        contentLike
+        'absolute top-2 left-2 px-2 text-sm flex gap-x-2 items-center dark:text-white',
```
chriswhong, line 26: Can we just replace `z-0` with `z-10` in the same string on line 26?

chriswhong (changes requested):

PR #1014, opened 2026-06-15
Compress share URLs with gzip + base64url

AndrewSepic:
## Summary

- Replaces the percent-encoded `data:application/json,...` payload with gzip-compressed, base64url-encoded data (`?data=gz:<base64url>`)
- Adds `app/lib/url_encoding.ts` with `gzipEncode`/`gzipDecode` helpers using the browser-native `CompressionStream` API (Chrome 80+, Firefox 113+, Safari 16.4+)
- Old `?data=data:application/json,...` links continue to work — the decoder in `url_api.tsx` falls through to the existing path
- Typical GeoJSON compresses ~90%, pushing the practical data limit from ~3 KB to ~20 KB within the same 8k-character URL budget
- Updates the Share button in the toolbar to use the Radix `Share2Icon` instead of the word "Share"
Screenshot 2026-06-15 at 11 44 11 AM

## Test plan

- [ ] Open the Share dialog with a small dataset — URL should start with `?data=gz:`
- [ ] Copy and open the share URL in a new tab — data should load and map should zoom to fit
- [ ] Test with a large dataset (~50 features) — should still fit within the URL limit where the old encoding would have hit the warning
- [ ] Confirm the "too large" error still appears for a very large dataset
- [ ] Confirm the toolbar Share button shows the icon (not the word "Share")
- [ ] Paste an old-format `?data=data:application/json,...` URL — should still load correctly


chriswhong (approved):
This is looking great and from my testing allows for sharing of geojson that would be in the low 800kb in file size.

PR #1015, opened 2026-06-17
Fix GA4 firing page_view on every map interaction

chriswhong:
## Summary

- GA4's Enhanced Measurement tracks `history.replaceState` calls as `page_view` events
- The map camera position is written to the `?map=` query param via `replaceState` on every pan/zoom, flooding GA4 with page views
- Fixes by setting `send_page_view: false` on the gtag config and manually firing a single `page_view` on load with a clean URL (no query params)

## Test plan

- [ ] Open the app and confirm a single `page_view` event fires on load (check GA4 DebugView or network requests to `google-analytics.com`)
- [ ] Pan/zoom the map and confirm no additional `page_view` events are fired

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,7 +20,14 @@
       window.dataLayer = window.dataLayer || [];
       function gtag(){dataLayer.push(arguments);}
       gtag('js', new Date());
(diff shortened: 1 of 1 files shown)


AndrewSepic (approved):
👍

PR #1019, opened 2026-06-25
Fix two-finger pan while drawing & Added panel resizing surface on mobile devices.

### LINKED ISSUE #974: Tablet functionality
Heyo,

This may be a misunderstanding of the new interface, but I use geojson to map with my iPad, and I find in the new version that when I am drawing a line string (or polygon, etc) I cannot use two fingers to navigate the map. Every interaction with the screen adds additional points, which is difficult when the shape I am drawing requires zooming in for details or On the old version, this works perfectly (well, sometimes the map gets locked on one point, but it reverts to correct behavior when I reload). Some of the new features look lovely, but without the ability to adjust the map while I draw on a tablet, geojson loses all functionality for me. 

Thanks much! Geojson is so useful for me, so I really hope this gets fixed!

### LINKED ISSUE #986: Panels are not resizeable/selectable on Mobile Devices
When using the new application on a mobile device (iOS Safari iPhone 16pro). You cannot select and change the size of the geojson panel, which limits your ability to interact with the map. The panel size selector utility is activated on hover, and so is not available via mobile devices.

In the image below, there is now way to increase the real estate of the map.
Image

AndrewSepic:
## Fix Two-finger pan while drawing (Reported in #974)

- On touch devices (iPad, iPhone), two-finger pan/zoom gestures during line or polygon drawing were being swallowed by the drawing handlers, preventing Mapbox GL from navigating the map
- Root cause: `touchstart` unconditionally called `e.preventDefault()` and `touchend` unconditionally called `handlers.click()`, regardless of how many fingers were involved
- Fix adds an `isMultiTouch` ref in both `useLineHandlers` and `usePolygonHandlers` — when two or more fingers are detected, touch events are passed through to Mapbox GL rather than treated as drawing actions

### How it works

- `touchstart`: if `touches.length > 1`, set `isMultiTouch = true` and return without calling `preventDefault()` — the map handles the gesture natively
- `touchmove`: bail out if multi-touch, so no phantom point updates during a pan
- `touchend`: skip `click` if `isMultiTouch` is set; reset the flag only once all fingers have lifted (handles the case where two fingers lift one at a time)

Closes #974 

## Fix Mobile Panel Resizability
- On a mobile device the horizontal panel at app bottom did not make a draggable surface visbille or available to click.  Forcing the mobile user to not be able to resize the panel, and providing much reduced map surface. 
- This adds CSS styles to ensure that the handle to resizing the panel is visible and can be used on Mobile devices.
Screenshot 2026-06-25 at 2 57 03 PM

Closes #986

## Remove LegendContainer & Small Design artifact.
- There is a very small DOM element visible below the Legend, which was seen and after review this can be removed.  I believe our refactor of the Placemark.io legend made this container un-neccesary.
Screenshot 2026-06-25 at 11 40 12 AM

## Test plan

- [x] On a touch device, enter line drawing mode, place a point, then use two fingers to pan/zoom — map should navigate without adding new points
- [x] Confirm single-finger tap still adds points correctly
- [x] Repeat for polygon drawing mode
- [x] Confirm Apple Pencil drawing still works (existing `force` check in polygon handler is preserved)
- [ ] Confirm a that a draggable handle is visible and can be used on small devices to change the size of the bottom panel.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Recent changes to the geojson.io application. Entries are grouped by ship date.
 
 ## 2026-06-25
 
+- **Fix 2 finger pan while in drawing mode:** Fix for users on touch devices, now 2 finger gestures while drawing is in process will make it to the map and will allow for panning.
+- **Mobile Panel Resizing:** On Mobile devices the bottom panel now presents clickable surface for resizing.
 - **Streamlined file import:** Files with unambiguous formats (GeoJSON, KML, GPX, shapefiles, etc.) now import directly without showing a dialog. The import modal only appears when user input is needed (CSV, XLS, coordinate strings). Drag-and-drop and the Import button now share identical logic, and `fitBounds` works correctly after drag-and-drop imports.
 
 ## 2026-06-24
--- a/app/components/legend.tsx
+++ b/app/components/legend.tsx
@@ -48,19 +48,6 @@ function LegendTitle({ title }: { title: string }) {
   );
 }
 
-function LegendContainer({ children }: { children: React.ReactNode }) {
-  return (
-    <div
-      className="space-y-1 text-xs
-      bg-white dark:bg-gray-900
-      dark:text-white
-      border border-gray-300 dark:border-black w-48 rounded-t"
-    >
-      {children}
-    </div>
-  );
-}
-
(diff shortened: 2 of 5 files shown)


chriswhong (approved):
LGTM, actually tested this with an iPad and Apple pencil and an iPhone, everything works as expected.

PR #1025, opened 2026-08-10
fix: use key to correctly detect undo/redo keybind on qwertz keyboards

### LINKED ISSUE #1024: Undo function is wired incorrectly on QWERTZ keyboard
My setup:
- MacOS 15.4.1
- Firefox 152.0
- German external keyboard (QWERTZ)
- German keyboard layout configured ("DE" in the menu bar)
- (Ctrl/Windows/Alt keys remapped to match Windows/Linux behaviour -- I don't think this is relevant, but mentioning anyway)

When I draw a feature and hit Ctrl+Z (the key between T and U on my keyboard), nothing happens. When I click Ctrl+Y though (the key next to X on my keyboard), the feature disappears. When I then hit Ctrl+Z again, the feature doesn't come back.

-> Undo is wired to shortcut where I would expect Redo. Redo doesn't work at all.

Some more observations:
- The behaviour is the same when using the keyboard built into the MacBook itself.
- When I type into the code textfield on the right, pressing Z on my keyboard does produce a Z as expected, and Y produces a Y as expected.
- Ctrl+S works normally (opens the export dialogue).

Let me know which other info you need to troubleshoot/recreate this.

Related to #673 I guess

Xstoudi:
Fixes #1024

Let me know if you think an other keybinds could actually need this change.

### DIFF
--- a/app/hooks/use_map_keybindings.ts
+++ b/app/hooks/use_map_keybindings.ts
@@ -41,6 +41,7 @@ export function useMapKeybindings() {
       historyControl('undo').catch((e) => captureException(e));
       return false;
     },
+    { useKey: true },
     [historyControl]
   );
 
@@ -49,6 +50,7 @@ export function useMapKeybindings() {
     (_e: KeyboardEvent) => {
       historyControl('redo').catch((e) => captureException(e));
     },
+    { useKey: true },
     [historyControl]
   );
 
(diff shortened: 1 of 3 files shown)


AndrewSepic (approved):
Thanks @Xstoudi  for the contribution 👍  

Closes #1024

